PR #2, opened 2015-09-29
Initial working version of UI

N3da:


jade/index.jade
```diff
@@ -0,0 +1,61 @@
...
+html(lang='en')
+  head
+    meta(charset='utf-8')
+    meta(name='description', content='meta desc')
+    meta(name='viewport', content='width=device-width')
+    title Avsc
+    link(rel='stylesheet', href='../stylesheets/main.css', type='text/css')
+  
+  body
+    script(src='../js/bundle.js')
```
mtth: In case you were wondering if it matters where `script` tags go, [this question](http://stackoverflow.com/q/436411/1062617) is a good read.

jade/index.jade
```diff
@@ -0,0 +1,61 @@
...
+    div.container
+      table
+        col.wide
+        col.narrow
+        col.wide
+        tr.input-row
+          td
+            div.outer
+              div
+                textarea#schema.textbox.verifiable
```
mtth: Browsers don't complain but it seems having `textarea` elements outside a `form` (or without a `form` attribute) is [illegal](http://stackoverflow.com/q/14528951/1062617)!

js/index.js
```diff
@@ -0,0 +1,128 @@
...
+        error_elem.removeClass('error');
+        error_elem.addClass('valid');
+        error_elem.text('Valid Schema!');
+      } catch (err) {
+        showError(error_elem, err);
+      }
+    }
+ 
+    $('#encode').click(function() {
+      var schema = readInput('#schema');
```
mtth: What do you think of saving the parsed schema into a variable, and using its current value when a decode/encode interaction comes in? (Only parsing it when the `#schema` element changes.)

js/index.js
```diff
@@ -0,0 +1,128 @@
...
+      var input = readInput('#input');
+      var p = avsc.parse(schema);
+      try {
+        var output = p.encode(input);
+        $('#output').val(bufferToStr(output));
+      }catch(err) {
+        showError($('#decoded-error'),err);
```
mtth: It doesn't look like these errors get cleared off? (This element and ones below seem to be missing some handling that `#schema` has.)

mtth:
This is awesome, thank you!

jade/index.jade
N3da: Great! Thank you!

jade/index.jade
N3da: :-/ huh!

mtth:
Thank you :).

PR #27, opened 2016-01-22
Update UI

N3da:


js/index.js
```diff
@@ -2,128 +2,798 @@
...
+        eventObj = utils.eventObj,
+        doneTypingInterval = 500; // wait for some time before processing user input.
+    
+    window.reverseIndexMap = [];  
 
-  require('jquery-ui');
-  window.avsc = avsc;
+    eventObj.on('schema-changed', function() {
+      runPreservingCursorPosition( 'schema', validateSchema);
+    }).on('input-changed', function() {
```
mtth: It might make the code a bit simpler to understand if we passed arguments to the handlers (instead of relying on globals). For example here `'input-changed'` could get the trimmed input.

js/index.js
```diff
@@ -2,128 +2,798 @@
...
+      clearTimeout(typingTimer);
+      typingTimer = setTimeout(function () {
+        if(updateContent(schemaElement)) {
+          eventObj.trigger('schema-changed');
+        }
+      }, doneTypingInterval);
+    }).on('keydown', function() {
+      clearTimeout(typingTimer);
+    }).on('click', function(event) {
+      if($(this).hasClass('-placeholder-')) {
```
mtth: I haven't looked into it but I've heard it's good practice to cache jQuery selectors (`$(this)` here).

js/index.js
```diff
@@ -2,128 +2,798 @@
...
+        highlight($(this)); 
+
+        /*So that the parent won't be highlighted (because we are using mouseover and not mouseenter)*/
+        event.stopPropagation(); 
+
+        var path = getPath($(this));
+        var position = findPositionOf(path);
+        highlight(position); 
+      } else 
+        console.log("No instrumented type found");
```
mtth: Braces?

js/index.js
```diff
@@ -2,128 +2,798 @@
...
+      console.log("uploading");
+      e.preventDefault();
+      $("#upload").trigger('click');
+      return false;
+    });
+
+    $('#upload').on("change", function(e) {
+      console.log("new file uploading");
+      var files = $('#upload')[0].files;
+      if(!!files && files.length > 0) {
```
mtth: `!!` is redundant here, since it is only used as a predicate. (Same in a few other places.)

js/index.js
```diff
@@ -2,128 +2,798 @@
...
+    *Read the text represented as space-seperated hex numbers in elementId
+    *and construct a Buffer object*/
+    function readBuffer(str) {
+      var rawInput = str || $.trim(outputElement.text());
+      var hexArray = rawInput.split(/[\s,]+/);
+      var i;
+      var size = hexArray.length;
+      var buffer = [];
+      for (i =0; i < size; i++){
+        buffer.push(new Buffer(hexArray[i], 'hex'));
```
mtth: Rather than split and combine (which requires the creation of multiple buffers), would it work to remove the commas and whitespace from the original input and pass the result directly?

js/index.js
N3da: Good point. This particular one I could actually remove. But I cached a few of other ones. (There are still some left but I think they are legitimately 'uncachable' as far as I could tell (e.g. for each div element that's being highlighted). I might be wrong though :)

js/index.js
N3da: foreach?

js/index.js
N3da: but... that would be too simple!:)

js/index.js
N3da: thanks :)

js/index.js
mtth: :trophy:

PR #49, opened 2016-05-02
Add infer and assemble pages 

N3da:


mtth:
Thank you @N3da, this is awesome!

PR #51, opened 2016-05-14
Enforce reading buffer as characters for browser readString

cosm1c:
I was attempting to use avsc.decode on a WebSocket arraybuffer after reading other types before on the same buffer and instead of getting the string values I was getting "123,456,789".

This small change fixes this by ensuring the bytes are read as characters.

### DIFF
--- a/etc/browser/lib/files.js
+++ b/etc/browser/lib/files.js
@@ -23,7 +23,7 @@
     if (this.pos > buf.length) {
       return;
     }
-    return this.buf.slice(pos, pos + len).toString();
+    return String.fromCharCode.apply(null, this.buf.slice(pos, pos + len));
   };
 
   Tap.prototype.writeString = function (s) {
--- a/etc/browser/lib/files.js
+++ b/etc/browser/lib/files.js
@@ -23,7 +23,7 @@
     if (this.pos > buf.length) {
       return;
     }
-    return String.fromCharCode.apply(null, this.buf.slice(pos, pos + len));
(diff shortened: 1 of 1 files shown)


mtth:
Thanks! Could you give an example of input that causes this (in particular how the `ArrayBuffer` is built)?

cosm1c:
The data was obtained from a WebSocket.onMessage (binaryType = 'arraybuffer') and wrapped in a Uint8Array.
ie:
var data = new Uint8Array(e.data);

Then data is used as the ArrayBuffer input to avsc (as the raw ArrayBuffer does not allow array index access).

I just updated the PR to support Safari and other browsers that don't provide an ArrayBuffer.slice method on the raw data in.

mtth:
I see. The problem here is that you are using a `Uint8Array` while all decoding methods expect [`Buffer`](https://github.com/feross/buffer) objects. Using your example but wrapping the `ArrayBuffer` into a `Buffer` instead gives correct results:

```javascript
// Some sample data (the Avro encoding of `"hi"`).
var ua = new Uint8Array([4, 104, 105]);
// The underlying `ArrayBuffer` (similar to what you are receiving from the WebSocket).
var ab = ua.buffer;
// Transform it to a `Buffer` (not `Uint8Array`).
var buf = new Buffer(ab);
// Now, encoding works as expected:
var type = avro.parse('string');
console.log(type.fromBuffer(buf)); // `"hi"`
```

Would this work for your use-case?

cosm1c:
Yes that solved the problem quite nicely - thank you!

mtth:
Glad to hear, @cosm1c!

PR #73, opened 2016-10-29
record type backward compability with default

bencebalogh:
The [avro specification about schema resolution](http://avro.apache.org/docs/1.8.1/spec.html#Schema+Resolution) says:
> if the reader's record schema has a field that contains a default value, and writer's schema does not have a field with the same name, then the reader should use the default value from its field.

More about this [here](http://docs.confluent.io/1.0.1/avro.html) with examples.

With the current version if I follow the examples from [above](http://docs.confluent.io/1.0.1/avro.html) I get a truncated buffer when calling `fromBuffer`.

Example with the current version:
```
'use strict';

const avro = require('avsc');

const schema1 = avro.parse({
    "namespace": "example.avro",
    "type": "record",
    "name": "user",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "favorite_number",
            "type": "int"
        }
    ]
});
const schema2 = avro.parse({
    "namespace": "example.avro",
    "type": "record",
    "name": "user",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "favorite_number",
            "type": "int"
        },
        {
            "name": "favorite_color",
            "type": "string",
            "default":  "green"
        },
        {
            "name": "happy",
            "type": "boolean",
            "default": true
        },
        {
            "name": "age",
            "type": "int",
            "default": 12
        },
        {
            "name": "data",
            "type": {
                "type": "map",
                "values": ["long", "string"]
            },
            "default": {
                "hah": 11111
            },
        }
    ]
});
const buf = schema1.toBuffer(message);

const val1 = schema1.fromBuffer(buf);
console.log(JSON.stringify(val1, 1, 2));

const val2 = schema2.fromBuffer(buf);
console.log(JSON.stringify(val2, 1, 2));

```

Examples with the changes:
```
'use strict';

const avro = require('avsc');

const schema1 = avro.parse({
    "namespace": "example.avro",
    "type": "record",
    "name": "user",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "favorite_number",
            "type": "int"
        }
    ]
});
const schema2 = avro.parse({
    "namespace": "example.avro",
    "type": "record",
    "name": "user",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "favorite_number",
            "type": "int"
        },
        {
            "name": "favorite_color",
            "type": "string",
            "default":  "green"
        },
        {
            "name": "happy",
            "type": "boolean",
            "default": true
        },
        {
            "name": "age",
            "type": "int",
            "default": 12
        },
        {
            "name": "data",
            "type": {
                "type": "map",
                "values": ["long", "string"]
            },
            "default": {
                "hah": 11111
            },
        }
    ]
});
const schema3 = avro.parse({
    "namespace": "example.avro",
    "type": "record",
    "name": "user",
    "fields": [
        {
            "name": "name",
            "type": "string"
        },
        {
            "name": "favorite_number",
            "type": "int"
        },
        {
            "name": "favorite_color",
            "type": "string",
            "default":  "green"
        },
        {
            "name": "happy",
            "type": "boolean"
        }
    ]
});

const message = {
    "name": "Joe",
    "favorite_number": 1
};

const buf = schema1.toBuffer(message);

const val1 = schema1.fromBuffer(buf);
console.log(JSON.stringify(val1, 1, 2));

const val2 = schema2.fromBuffer(buf);
console.log(JSON.stringify(val2, 1, 2));

try {
    schema3.fromBuffer(buf);
} catch(e) {
    console.log('expect to throw');
}
```

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -1834,7 +1834,7 @@ util.inherits(RecordType, Type);
 RecordType.prototype.fromBuffer = function (buf, resolver, noCheck) {
   var tap = new Tap(buf);
   var val = readValue(this, tap, resolver, noCheck);
-  let tapPos = tap.pos;
+  var tapPos = tap.pos;
   if (!tap.isValid()) {
     // Find the first field with default value which is only followed by fields with default values
     // or it's the last field in the Record
--- a/lib/types.js
+++ b/lib/types.js
@@ -1831,6 +1831,73 @@ function RecordType(attrs, opts) {
 }
 util.inherits(RecordType, Type);
 
+RecordType.prototype.fromBuffer = function (buf, resolver, noCheck) {
+  var tap = new Tap(buf);
+  var val = readValue(this, tap, resolver, noCheck);
+  let tapPos = tap.pos;
+  if (!tap.isValid()) {
+    // Find the first field with default value which is only followed by fields with default values
+    // or it's the last field in the Record
+    var firstDefaultIndex = this._fields.indexOf(this._fields.find(checkNextField.bind(this)));
+    
+    function checkNextField(field, index) {
+      if (field.getDefault() !== undefined) {
+        if (index == this._fields.length - 1) {
+          return true;
+        }
+        
+        var nextField = this._fields[index + 1];
+        return checkNextField.bind(this)(nextField, index + 1);
+      }
+      
+      return false;
+    }
+    
+    var validWithDefaults = false;
+    
+    if (firstDefaultIndex != -1) {
+      // Encode all default values after index and check if the buffer is valid
+      validWithDefaults = checkFields.bind(this)(firstDefaultIndex);
+      
+      function checkFields(index) {
+        var valid;
+        
+        var checkFields = this._fields.slice(index, this._fields.length);
+        var offset = 0;
+        
+        checkFields.map(function(c) {
+          c._type._write(tap, c.getDefault());
+        
+          offset += c._type.encode(c.getDefault(), buf, tap.pos);
+        });
+        
+        valid = tap.pos <= (-1 * offset) + tap.buf.length;
+        
+        if (!valid && index < this._fields.length - 1) {
+          checkFields.bind(this)(indexOfDefField);
+        }
+        
+        if (valid) {
+          // If it's valid set default values on the return object
+          checkFields.map(function(field) {
+            val[field._name] = field.getDefault();
+          });
+        }
+        
+        return valid;
+      }
+    }
+    
+    if (!validWithDefaults) {
+      throw new Error('truncated buffer');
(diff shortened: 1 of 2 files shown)


bencebalogh:
The PR is not final yet, there are cases which are not correct: the record in the writer's schema has a last default key, reader has that key plus more extra defaults, the last default key in the writer's schema will be overriden to default value as well, which is incorrect.

mtth:
To read data from another schema, you need to pass in a [`resolver`](https://github.com/mtth/avsc/wiki/API#typecreateresolverwritertype) to `fromBuffer`. In your particular example:

```javascript
const avro = require('avsc');

const schema1 = avro.parse(/* ... */);
const schema2 = avro.parse(/* ... */);

// Instantiate a resolver that will let schema2 read schema1-encoded buffers.
const resolver = schema2.createResolver(schema1);

// Encode a message using schema1.
const buf = schema1.toBuffer({
    "name": "Joe",
    "favorite_number": 1
});

// Decode it using schema2, with the above resolver.
const resolvedMessage = schema2.fromBuffer(buf, resolver);
// Where the resolved message is as expected:
// {
//   "name": "Joe",
//   "favorite_number": 1,
//   "favorite_color": "green",
//   "happy": true,
//   "age": 12,
//   "data": {
//     "hah": 11111
//   }
// }
```

There are multiple reasons to create the resolver ahead of time:

+ `fromBuffer` doesn't have enough information to correctly perform all the operations explained in the spec unless we also pass in the original schema (for example remapping fields if their order or names changed).
+ It's an expensive operation (similar to parsing a schema), better done ahead of time once.
+ This lets us validate whether two schemas are compatible with only the schemas.

You can read more about schema evolution on the [advanced usage](https://github.com/mtth/avsc/wiki/Advanced-usage#schema-evolution) page. Let me know if anything doesn't make sense.

gogoout:
Hi @mtth ,
I just want to put my 2 cents on allow this behaviour. But according to the spec, this behaviour should be allowed.

I don't know about the limitation you mentioned above, correct me if I'm wrong but the java implementation does support this behaviour, see [here](https://github.com/apache/avro/blob/master/lang/java/avro/src/test/java/org/apache/avro/TestReadingWritingDataInEvolvedSchemas.java#L373)

As for the performance issue, I understand this will have a performance hurt. But by requiring the writer schema, user have to store all the schema a value have used before. For manual defined schema, it maybe just a bit loss in usability. In our case, we generate schema automatically based on upstream schema(eg. db schema) changes, and it's just impossible to handle this correct.

PR #78, opened 2016-12-06
Patch: Don't try to copy a value if it doesn't exist

adambraimbridge:
When reproducing the solution described in [this stackoverflow issue](http://stackoverflow.com/questions/33236284/how-to-convert-a-json-object-into-avro-object-if-avro-schema-contains-union-in-i/33622239#33622239), I encountered a bug; specifically when my schema contains a child record — which my data does not contain (as it's meant to be optional)

The `type` I used to reproduce the bug:
```javascript
const type = avsc.parse(
	{
		'type':'record',
		'name':'DataFlowEntity_Level0',
		'fields': [
			{'name':'level0_null','type':['null','string', 'int']},
			{'name':'level0_string','type':['null','string', 'int']},
			{'name':'level0_int','type':['null','string', 'int']},
			{'name':'level0_boolean','type':['null','string', 'int', 'boolean']},
			{'name':'level0_record', type: [
				'null',
				{
					'type': 'record',
					'name': 'DataFlowEntity_Level1',
					'fields': [
						{'name':'level1_null','type':['null','string', 'int']},
						{'name':'level1_string','type':['null','string', 'int']},
						{'name':'level1_int','type':['null','string', 'int']},
						{'name':'level1_boolean','type':['null','string', 'int', 'boolean']},
					]
				}
			]}
		]
	},
	{
		wrapUnions: true
	}
);
```

The data for the test:
```javascript
const invalidRecord_withoutLevel1 = {
	'level0_null': null,
	'level0_string': 'example string',
	'level0_int': 123456,
	'level0_boolean': true,
};

const invalidRecord_withLevel1 = Object.assign({
	'level0_record': {
		'level1_null': null,
		'level1_string': 'example string',
		'level1_int': 123456,
		'level1_boolean': true,
	}
}, invalidRecord_withoutLevel1);
```

Running this code causes an error:
```javascript
const clonedRecord = type.clone(invalidRecord_withoutLevel1, { wrapUnions: true });
```

... which this PR attempts to patch in  in [avsc/lib/types.js](https://github.com/mtth/avsc/blob/master/lib/types.js#L2127).

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -2115,6 +2115,7 @@ RecordType.prototype._checkFields = function (obj) {
 
 RecordType.prototype._copy = function (val, opts) {
   // jshint -W058
+  val = val || [];
   var hook = opts && opts.fieldHook;
   var values = [undefined];
   var i, l, field, value;
--- a/lib/types.js
+++ b/lib/types.js
@@ -2123,7 +2123,7 @@ RecordType.prototype._copy = function (val, opts) {
     value = val[field._name];
     if (value === undefined && field.hasOwnProperty('getDefault')) {
       value = field.getDefault();
-    } else if(val[field._name]) {
+    } else if(val[field._name] !== undefined) {
       value = field._type._copy(val[field._name], opts);
     }
     if (hook) {
--- a/lib/types.js
+++ b/lib/types.js
@@ -2123,7 +2123,7 @@ RecordType.prototype._copy = function (val, opts) {
     value = val[field._name];
     if (value === undefined && field.hasOwnProperty('getDefault')) {
       value = field.getDefault();
-    } else {
+    } else if(val[field._name]) {
       value = field._type._copy(val[field._name], opts);
     }
     if (hook) {


mtth:
Hi Adam, the code is working as intended. The fields in your schema are nullable but not _optional_: to be the latter, they would need a default. There are therefore two ways of fixing things here:

+ Adding a `null` default to all fields declarations:

```javascript
const type = avsc.parse({
  'type': 'record',
  'name': 'DataFlowEntity_Level0',
  'fields': [
    {'name':'level0_null', 'type':['null','string', 'int'], 'default': null},
    // Same for the other fields (including `level0_record`)...
  ]
}, {wrapUnions: true});
```

+ Making the field `null` explicitly:

```javascript
const invalidRecord_withoutLevel1 = {
  'level0_null': null,
  'level0_string': 'example string',
  'level0_int': 123456,
  'level0_boolean': true,
  'level0_record': null
};
```

Hopefully this makes sense, unions can be unintuitive.

adambraimbridge:
You're definitely right, Matthieu, unions sure can be unintuitive. Your suggestion to add a null default to all field declarations has been successful for me. 

Many sincere thanks! Adam

adambraimbridge:
Hello again, Matthieu.

I've found a better case where I'm encountering that pesky `TypeError: Cannot read property '[x]' of undefined` error when running `type.clone()`.

All of the nullable properties are working as expected (thanks again for your advice), but in this case I've added a new, mandatory record. 

```javascript
{
	'type':'record',
	'name':'DataFlowEntity_Level0',
	'fields': [
		{'default':null,'name':'level0_null','type':['null','string', 'int']},
		{'default':null,'name':'level0_object','type':['null','string', 'int']},
		{'default':null,'name':'level0_string','type':['null','string', 'int']},
		{'default':null,'name':'level0_int','type':['null','string', 'int']},
		{'default':null,'name':'level0_boolean','type':['null','string', 'int', 'boolean']},
		{'default':null,'name':'level0_record', type: [
			'null',
			{
				'type': 'record',
				'name': 'DataFlowEntity_Level1',
				'fields': [
					{'default':null,'name':'level1_null','type':['null','string', 'int']},
					{'default':null,'name':'level1_string','type':['null','string', 'int']},
					{'default':null,'name':'level1_int','type':['null','string', 'int']},
					{'default':null,'name':'level1_boolean','type':['null','string', 'int', 'boolean']},
				]
			}
		]},
		{
			'name':'level0_record_mandatory', type: {
				'type': 'record',
				'name': 'DataFlowEntity_Level1_Mandatory',
				'fields': [
					{'name':'level1_string','type': 'string'},
				]
			}
		},
	]
};
``` 
_Above: Note that `level0_record_mandatory` isn't nullable nor optional._ 

Running the same code throws the error: 
```javascript
const clonedRecord = type.clone(invalidRecord_withoutLevel1, { wrapUnions: true });
```
I'm using the original version of `invalidRecord_withoutLevel1`, which doesn't  include `level0_record_mandatory`.

My expectation would be to receive no errors when calling `type.clone()`. And then, when calling `type.isValid(clonedRecord)`, I'd expect a validation error along the lines of "missing record of type: DataFlowEntity_Level1_Mandatory". 

The suggested patch _ought_ to do the trick, but I could be on the wrong track, as before. Any advice is greatly appreciated.

Thanks, Adam

mtth:
I see, `clone` does currently expect valid values (union-wrapping and buffer-coercion aside, and only if these options are active). Enabling better validation might be a reasonable use-case for skipping missing fields though. I wouldn't enable it by default but we could support it via a new flag, say `skipMissingFields`.

Your PR is a good start, you'll just need to add the option in `clone`'s [definition](https://github.com/mtth/avsc/blob/master/lib/types.js#L345) and toggle behavior accordingly. Or I can take care of it, let me know.

adambraimbridge:
Thank you again. As luck has it, I'm off travelling for a week, and likely won't have much computer time. So if you can take care of it, I'd sincere;y appreciate it. Otherwise I'll make a fresh PR, with the `skipMissingFields` option, in a week from now.

Cheers, Adam.

mtth:
No problem, just released `4.1.11` with the new option. Happy travels!

adambraimbridge:
I didn't have a chance to thank you earlier, @mtth. Much appreciated, thanks. Adam

PR #89, opened 2017-02-20
call `Error.captureStackTrace` if possible

gurpreetatwal:
The constructors for schema defined errors now call `Error.captureStackTrace`
if it is available. As of today `Error.captureStackTrace` is only
available in V8 environments (nodejs & Chrome).

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -111,6 +111,7 @@ function Type(schema, opts) {
 Type.forSchema = function (schema, opts) {
   opts = opts || {};
   opts.registry = opts.registry || {};
+  opts.noStackTraces = opts.noStackTraces !== undefined ? opts.noStackTraces : false;
 
   var UnionType = (function (wrapUnions) {
     if (wrapUnions === true) {
@@ -2071,6 +2072,7 @@ function RecordType(schema, opts) {
   }, this));
   this._branchConstructor = this._createBranchConstructor();
   this._isError = schema.type === 'error';
+  this._noStackTraces = opts.noStackTraces;
   this.recordConstructor = this._createConstructor();
   this._read = this._createReader();
   this._skip = this._createSkipper();
--- a/lib/types.js
+++ b/lib/types.js
@@ -2092,31 +2092,42 @@ RecordType.prototype._createConstructor = function (errorStackTraces) {
   // jshint -W054
   var outerArgs = [];
   var innerArgs = [];
-  var errorStackTraces = errorStackTraces === undefined ? false : errorStackTraces;
   var ds = []; // Defaults.
   var innerBody = '';
-  var i, l, field, name, defaultValue;
+  var errorStackTraces = errorStackTraces === undefined ? false : errorStackTraces;
+  var i, l, field, name, getDefault, defaultValue;
   for (i = 0, l = this.fields.length; i < l; i++) {
     field = this.fields[i];
-    defaultValue = field.defaultValue;
+    getDefault = field.defaultValue;
+    defaultValue = getDefault();
     name = field.name;
     innerArgs.push('v' + i);
     innerBody += '  ';
-    if (defaultValue() === undefined) {
+
+    // Call Error.captureStackTrace only if record is of type error and has a
+    // a string "stack" field
+    // istanbul ignore else
+    if (defaultValue === undefined && this._isError && errorStackTraces &&
+        name === 'stack' && field.type instanceof StringType &&
+        typeof Error.captureStackTrace === 'function') {
+
+      defaultValue = true;
+      getDefault = function(ctx) {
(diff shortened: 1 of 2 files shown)


test/test_types.js
```diff
@@ -2130,6 +2130,7 @@ suite('types', function () {
       var E = t.getRecordConstructor();
       var err = new E('MyError');
       assert(err instanceof Error);
+      assert(err.stack, 'Error.captureStackTrace was not called');
```
gurpreetatwal: ah yes, I (incorrectly) thought that the tests only ran on Node

lib/types.js
```diff
@@ -2094,6 +2094,10 @@ RecordType.prototype._createConstructor = function () {
   var innerArgs = [];
   var ds = []; // Defaults.
   var innerBody = '';
+  // `Error.captureStackTrace` is only available in V8 environments
+  if (this._isError && Error.captureStackTrace) {
+    innerBody = 'Error.captureStackTrace(this);\n';
```
gurpreetatwal: passing `this.constructor` right? I normally do that forgot here somehow, I'll add it in

lib/types.js
```diff
@@ -111,6 +111,7 @@ function Type(schema, opts) {
 Type.forSchema = function (schema, opts) {
   opts = opts || {};
   opts.registry = opts.registry || {};
+  opts.noStackTraces = opts.noStackTraces !== undefined ? opts.noStackTraces : false;
```
gurpreetatwal: Wasn't too sure where exactly to place this code, should the defaulting happen here on closer to where it used? I couldn't infer your style from the code.

lib/types.js
```diff
@@ -2071,6 +2072,7 @@ function RecordType(schema, opts) {
   }, this));
   this._branchConstructor = this._createBranchConstructor();
   this._isError = schema.type === 'error';
+  this._noStackTraces = opts.noStackTraces;
```
gurpreetatwal: should this happen in the `RecordType` constructor or in the `Type` constructor?

test/test_types.js
```diff
@@ -2130,6 +2130,52 @@ suite('types', function () {
       var E = t.getRecordConstructor();
       var err = new E('MyError');
       assert(err instanceof Error);
+      if (typeof Error.captureStackTrace === 'function') {
+        assert(err.stack, 'Error.captureStackTrace was not called');
+      }
+    });
+
+    test('error type - stack is not overridden', function() {
```
gurpreetatwal: how do you feel about moving the error tests into their own subsuite?

lib/types.js
mtth: I usually defer until later, except where recursion makes it easier to do it upfront. In this case I'd simply pass the options through, not mentioning it here (see below).

lib/types.js
```diff
@@ -2071,6 +2072,10 @@ function RecordType(schema, opts) {
   }, this));
   this._branchConstructor = this._createBranchConstructor();
   this._isError = schema.type === 'error';
+
+  // `Error.captureStackTrace` is only available in V8 environments
+  this._showStack = !opts.noStackTraces && typeof Error.captureStackTrace === 'function';
```
mtth: I'd remove this and just pass in the option to `this._createConstructor` (see other comment above). The comment should go in the conditional block where `Error.captureStackTrace` is called.

lib/types.js
```diff
@@ -2111,6 +2116,12 @@ RecordType.prototype._createConstructor = function () {
       ds.push(defaultValue);
     }
   }
+  // istanbul ignore else
+  if (this._isError && this._showStack) {
+    innerBody += 'if (!this.stack) {\n';
```
mtth: The check should be `this.stack === undefined`. (Let's also indent the statements by two extra spaces to align them with the rest. I'd also just put them all one a single line, similar to the others above, and add a new line at the very end.)

test/test_types.js
mtth: We could if errors end up differing significantly from records; for now I'd say let's stick with the current hierarchy.

test/test_types.js
```diff
@@ -2130,6 +2130,52 @@ suite('types', function () {
...
+        fields: [
+          {name: 'name', type: 'string'},
+          {name: 'stack', type: 'string'},
+        ]
+      });
+
+      var E = t.getRecordConstructor();
+      var err = new E('MyError', 'my amazing stack');
+      assert(err instanceof Error);
+      assert(err.stack === 'my amazing stack');
```
mtth: Can you add a check for when no stack value is passed in as argument (ensuring that the stack is populated)?

mtth:
Merged, thanks @gurpreetatwal!

PR #96, opened 2017-04-06
Windows compat: use path.sep instead of '/' in gereric parse().

simonbuchan:
This gets the tests passing on windows, and should only affect people on that.

I'd think that regex checks would be better, since JSON and IDL are pretty unambiguous (though hardly trivial), but I suppose you *could* name a file `{"type": "fixed", "size": 100}` in Linux :)

### DIFF
--- a/lib/specs.js
+++ b/lib/specs.js
@@ -168,10 +168,11 @@ function assembleProtocol(fpath, opts, cb) {
  *
  * The parsing logic is as follows:
  *
- * + If `str` contains a `/` and is a path to an existing file, it will first
- *   be read as JSON, then as an IDL specification if JSON parsing failed. If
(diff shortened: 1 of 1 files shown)


simonbuchan:
CI fails cause the `doc` submodule is committed with missing commit `51f137d` - should I push the current `master` (`4645327`) to get it passing for now?

mtth:
Thanks for the PR @simonbuchan, this looks good. I pushed a couple `doc` commits and synced the submodule; can you rebase and check that the CI tests pass?

simonbuchan:
OK, done. (sorry for the delay - I just got into work 😴)

mtth:
No problem, thanks for the fix! Will release it in `5.0.1` in a few minutes.

PR #101, opened 2017-05-10
createFileAppender with flush support

ronaldocpontes:
Hi @mtth 

Sending a new file encoder with support to append new values to existing avro files and also flushing.

We need these functionalities to make writing events to avro files on http servers more reliable and with more writing guarantees in case of server failures.

Regards,
Ronaldo

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,27 @@ const avro = require('avsc');
   ];
   ```
 
++ Use a [writable stream][writable-stream] to write or append encoded values to an Avro
+  container file:
+
+  ```javascript
+  const type = avro.Type.forValue({
+    city: 'Cambridge',
+    zipCodes: ['02138', '02139'],
+    visits: 2
+  });
+
+  const opts =  {};
+  
+  // Use this option if you like to flush values as soon as possible using 1 block per value
+  // (better for continuous writing certaintiy at the cost of performance and larger file sizes)
+  opts.writeFlushed = true;
+
+  const encoder = avro.createFileAppender('./values.avro', type, opts);
+  encoder.write({city: 'Seattle', zipCodes: ['98101'], visits: 3});
+  encoder.end({city: 'New York', zipCodes: ['10001'], visits: 47});
+  ```
+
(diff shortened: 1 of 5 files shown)


mtth:
Hi @ronaldocpontes, thanks for the PR. I'll take a look this weekend.

mtth:
Hey @ronaldocpontes, I looked through the changes and I think it would make more sense to make them available as a separate package.

+ `createFileAppender` seems useful but not generic enough to offset the two new dependencies. It also uses synchronous APIs (`mkdirp.sync` and `fs.openSync`) unlike `createFileDecoder` and `createFileEncoder` which are asynchronous.
+ `writeFlushed` actually promotes an anti-pattern. In most cases, each encoded value will be small and the overhead of creating a block per value prohibitive (both in terms of performance and size, especially if compression is enabled). If you have a use-case for immediately emitting encoded values, I would instead recommend implementing it as a separate encoder.

I could see both being included in a server-side-only Avro-file utilities library, which could then grow to include related functionality. For example, there was recently a question (https://github.com/mtth/avsc/issues/104) about parallel writes to Avro files and such a library could be a good place to provide tools for it.

ronaldocpontes:
Hey @mtth I am just a little worried on providing the functionality in another package / fork as I won't be able to maintain it and keep it updated with the latest changes from your repo.

Only referencing your package is also not ideal since I would have to rewrite the internals of the BlockEncoder and also the functionality wouldn't be available to the community at large. 

Not making createFileAppender sync creates additional complexities serverside, specially during the rolling of files where writes would need to be buffered elsewhere until the new file is opened (otherwise these writes will be silently lost). Maybe renaming the constructor to 
createFileAppenderSync would be more clear? 

The current implementation of BlockEncoder will also silently loose all writes in both the stream and the tap buffers if the server breaks or the end() method is not called before exit.

As these new functionalities are optional by the users of the package, do you think it would be too bad to keep them in so they can be updated with the rest of the package? 

If so, do you know any better ways to append to existing files and guarantee writes are flushed to permanent storage? 

Best 
Ronaldo

Issue #105, opened 2017-05-14
LogicalTypes in BlockDecoder


rgarza:
BlockDecoder is not using the logicalTypes defined for the type, add an option to pass in the logical Types.

mtth:
Hey @rgarza, it's actually supported via the [`parseHook`](https://github.com/mtth/avsc/wiki/API#class-blockdecoderopts) option (see also the example a little below; this lets us decouple the decoder's options from `Type.forSchema`'s):

```javascript
const decoder = new avro.streams.BlockDecoder({
  parseHook: (schema) => {
    return avro.Type.forSchema(schema, {logicalTypes});
  }
});
```

rgarza:
Ahh, makes sense, thank you

PR #106, opened 2017-05-14
use logical types in BlockDecoder

rgarza:
PR for issue #105.
Adds test to BlockEncoder to use logical types.
Adds option in BlockEncoder to receive the logicalTypes.


mtth:
Thanks for the PR, the functionality is already available actually! (See https://github.com/mtth/avsc/issues/105.)

PR #124, opened 2017-08-26
resolved issue causing downstream writables not firing finish events …

perryao:
…if an empty stream is piped into a BlockEncoder. 

@mtth When piping an empty read stream to an avsc BlockEncoder (e.g. a stream that only pushes null), and then piping to a writeStream (like from fs.createWriteStream), listening to `finish` events on the writeStream or any of the stream instances in that chain is not possible. 

I'm not 100% certain that this change fits conceptually, but I know it solves this issue I'm experiencing. Would love to chat through it to make sure this makes sense. If you comment out my change and run the test, you should see the issue reproduced.

### DIFF
--- a/lib/containers.js
+++ b/lib/containers.js
@@ -392,6 +392,11 @@ function BlockEncoder(schema, opts) {
     this._finished = true;
     if (this._blockCount) {
       this._flushChunk();
+    } else if (this._needPush) {
+        // if we finished but _needPush, we
+        // need to push null for downstream
+        // consumers
+        this.push(null);
     }
   });
 }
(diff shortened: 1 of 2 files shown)


mtth:
Thanks @perryao, this was definitely an oversight. Your change is a great start, we just need to be careful with pending asynchronous block compression. While I was looking at the code, I also noticed that we could improve the way we're emitting headers. I'll build on your commit to include all this.

mtth:
I've updated and released your fix in `5.0.5`; please let me know if anything isn't working. Thank you for the PR!

perryao:
Will do! Thanks for the merge @mtth. Love all the changes you made. Looks great

PR #134, opened 2017-11-06
Add (incomplete) Typescript definitions

andrew8er:
These are probably wrong and/or incomplete. They may serve as a starter though.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -95,7 +95,6 @@ export namespace Service {
 
   interface ClientOptions {
     buffering: boolean;
-    cache: any;
     channelPolicy: any;
     strictTypes: boolean;
     timeout: number;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,0 +1,151 @@
+import * as stream from 'stream';
+import { EventEmitter } from 'events'
+
+
+export type Callback<V, Err = any> = (err: Err, value: V) => void;
+
+export type CodecTransformer = (buffer: Buffer, callback: () => void) => Buffer; // TODO
+
+export interface CodecOptions {
+	deflate: CodecTransformer;
+	snappy: CodecTransformer;
+}
+
+export interface Decoder {
+	on(type: 'metadata', callback: (type: Type.Type) => void): this;
+	on(type: 'data', callback: (value: object) => void): this;
+}
+
+export interface Encoder {
+	// TODO
+}
+
+export interface ReaderOptions {
+	// TODO
+}
+
+interface AssembleOptions {
+	importHook: (filePath: string, type: 'idl', callback: Callback<object>) => void;
+}
+
+export function assembleProtocol(filePath: string, opts: AssembleOptions, callback: Callback<object>): void;
+export function assembleProtocol(filePath: string, callback: Callback<object>): void;
+export function createFileDecoder(fileName: string, codecs?: CodecOptions): Decoder;
+export function createFileEncoder(filePath: string, schema: any, options?: any): Encoder;
+export function discoverProtocol(transport: Service.Transport, options: any, callback: Callback<any>): void;
+export function discoverProtocol(transport: Service.Transport, callback: Callback<any>): void;
+export function extractFileHeader(filePath: string, options?: any): void;
+export function parse(schemaOrProtocolIdl: string, options?: any): Service.Protocol | Type.Type; // TODO
+export function readProtocol(protocolIdl: string, options?: ReaderOptions): Service.Protocol;
+export function readSchema(schemaIdl: string, options?: ReaderOptions): Type.Schema;
+// TODO streams
+// TODO types
+
+export namespace Type {
+	type Schema = string;
+
+	interface Type {
+		// TODO clone(val, opts)
+		// TODO compare
+		// TODO compareBuffers(buf1, buf2)
+		// TODO createResolver(type, opts)
+		// TODO decode(buf, pos, resolver)
+		// TODO encode(val, buf, pos)
+		// TODO equals(type)
+		// TODO fingerprint(algorithm)
+		fromBuffer(buffer: Buffer, resolver: any, noCheck: boolean): Type; // TODO
+		// TODO fromString(str)
+		// TODO inspect()
+		// TODO isValid(val, opts)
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -0,0 +1,151 @@
+import * as stream from 'stream';
+import { EventEmitter } from 'events'
+
+
+export type Callback<V, Err = any> = (err: Err, value: V) => void;
+
+export type CodecTransformer = (buffer: Buffer, callback: () => void) => Buffer; // TODO
+
+export interface CodecOptions {
+	deflate: CodecTransformer;
```
mtth: Is 8 spaces per tab a strong convention for typings? If not, can we switch to 2 for consistency with the other files?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
+import * as stream from 'stream';
+import { EventEmitter } from 'events'
+
+
+export type Callback<V, Err = any> = (err: Err, value: V) => void;
+
+export type CodecTransformer = (buffer: Buffer, callback: () => void) => Buffer; // TODO
```
mtth: It'd probably be easier to extend if we omit partial types entirely. Can you remove this and all other non-complete types in the file?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export interface CodecOptions {
+	deflate: CodecTransformer;
+	snappy: CodecTransformer;
+}
+
+export interface Decoder {
+	on(type: 'metadata', callback: (type: Type.Type) => void): this;
```
mtth: Does this mean that TypeScript is able to support different callback types for different events?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+export function assembleProtocol(filePath: string, callback: Callback<object>): void;
+export function createFileDecoder(fileName: string, codecs?: CodecOptions): Decoder;
+export function createFileEncoder(filePath: string, schema: any, options?: any): Encoder;
+export function discoverProtocol(transport: Service.Transport, options: any, callback: Callback<any>): void;
+export function discoverProtocol(transport: Service.Transport, callback: Callback<any>): void;
+export function extractFileHeader(filePath: string, options?: any): void;
+export function parse(schemaOrProtocolIdl: string, options?: any): Service.Protocol | Type.Type; // TODO
+export function readProtocol(protocolIdl: string, options?: ReaderOptions): Service.Protocol;
+export function readSchema(schemaIdl: string, options?: ReaderOptions): Type.Schema;
+// TODO streams
```
mtth: Unless you're certain that they are exhaustive, can we remove the commented `TODO`s?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+		readonly server: Server;
+		readonly destroyed: boolean;
+		readonly draining: boolean;
+		readonly pending: number;
+		destroy(noWait: boolean): void;
+	}
+
+	interface ClientOptions {
+		buffering: boolean;
+		cache: any;
```
mtth: The `cache` option is deprecated, can we remove it?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+	interface ServerOptions {
+		objectMode: boolean;
+	}
+
+	interface Service {
+		createClient(options?: ClientOptions): Client;
+		createServer(options?: ServerOptions): Server;
+		message(name: string): any;
+		type(name: string): any;
+		inspect(name: string): any;
```
mtth: I'm not sure it's useful to include this method, it's an override just for pretty-printing in the REPL.

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+		activeChannels(): ClientChannel[];
+		createChannel(transport: Transport, options?: ChannelCreateOptions): ClientChannel;
+		destroyChannels(options?: ChannelDestroyOptions): void;
+		emitMessage<T>(name: string, req: any, options?: any, callback?: Callback<T>): void // TODO
+		remoteProtocols(): Protocol[];
+		use(...args: any[]): this;
+	}
+
+	function compatible(client: Client, server: Server): boolean;
+	function forProtocol(protocol: Protocol): Service;
```
mtth: This might be too restrictive. It's sometimes useful to instantiate services from inlined JavaScript objects, which I think this would prevent, right?

types/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+export function discoverProtocol(transport: Service.Transport, callback: Callback<any>): void;
+export function extractFileHeader(filePath: string, options?: any): void;
+export function parse(schemaOrProtocolIdl: string, options?: any): Service.Protocol | Type.Type; // TODO
+export function readProtocol(protocolIdl: string, options?: ReaderOptions): Service.Protocol;
+export function readSchema(schemaIdl: string, options?: ReaderOptions): Type.Schema;
+// TODO streams
+// TODO types
+
+export namespace Type {
+	type Schema = string;
```
mtth: Schemas can also be JavaScript objects (for example `{type: 'map', values: 'int'}`).

types/index.d.ts
andrew8er: The `CodecTransformer` definition is almost correct, as far as I can see it, only the callback definition might be lacking an argument. I think, the definition has a certain value, because programmers will provide those functions and thus can be sure they are defined correctly.

types/index.d.ts
andrew8er: I have no strong feelings on file indentations, although I think that two spaces provide not enough visual contrast. I will  provide an extra commit.

types/index.d.ts
andrew8er: Yes it does!

types/index.d.ts
andrew8er: I am actually not sure if these are exhaustive, hence the TODO…

types/index.d.ts
andrew8er: The Js object could also be completely typechecked. Have a look at https://github.com/joewood/avro-typescript/blob/master/src/model.ts

Maybe we could extract a common Ts definition for Avro schemas for use in booth projects.

PR #137, opened 2017-11-14
Allows dashes in name patterns

joaomilho:


mtth:
Unfortunately dashes are illegal according to [Avro's specification](https://avro.apache.org/docs/1.8.0/spec.html#names) so this can't be merged. If there is sufficient interest, we could add a more general extension with custom name validation logic (extending the "anonymous types" one) which would cover this use-case. Let me know what you think.

joaomilho:
We were trying to follow Confluent's subject naming conventions (https://docs.confluent.io/current/schema-registry/docs/api.html#subjects) and taking the namespace to set it. So it's our misunderstanding :D

PR #145, opened 2018-01-09
Add more details to typescript annotation file, rearrange it.

TysonAndre:
Many of the types were generated by (a patched version of) dts-gen.
Because of that, many methods are vague and say `any`.

Rearrange the layout of the declarations to match what dts-gen emitted
(By dts-gen's design, this should match with the signature of `require('avsc')`)

- Keep the precise annotations that were already made, moving some.
- Make interfaces into classes where they correspond to an actual JS
  class (i.e. has function and function prototype)

A TODO was added to make places using types of `any` specific later.
I'm not familiar enough with this project to be certain of the intent,
and adding the methods/classes is a large enough PR on its own.

Add types/ folder to the package for `tsc`.

- This is necessary for that folder to be included in the package
  published to NPM

The namespaces and classes were generated by `dts-gen`.
Many `any` types are left over, I added a TODO to finish those later

I ran `tsc --strict` in a project that had a dependency of `avsc` to test this.
(The project uses only a few parts of avsc)

- Analysis of avsc.Type.forSchema worked as expected in a project that used
  a patched version of node_modules/avsc, with types/ subfolder added.
  (And toBuffer).
  (So, I know it is valid typescript)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2,6 +2,7 @@
 // TODO: Wherever the type is just `any`, it was probably generated automatically.
 //       Either finish documenting the type signature or document why `any` is appropriate.
 // TODO: Wherever the argument names are just `args: any`, it was probably generated from the signature of util.deprecate. Fix argument counts and types.
+// NOTE: This does not contain entries for functions available in the browser (functions/methods in etc/browser)
 
 import * as stream from 'stream';
 import { EventEmitter } from 'events'
@@ -39,7 +40,7 @@ export function assembleProtocol(filePath: string, opts: Partial<AssembleOptions
 export function assembleProtocol(filePath: string, callback: Callback<object>): void;
 export function combine(args: any): any;
 export function createFileDecoder(fileName: string, codecs?: Partial<CodecOptions>): Decoder;
-export function createFileEncoder(filePath: string, schema: any, options?: any): Encoder;
+export function createFileEncoder(filePath: string, schema: Schema, options?: any): Encoder;
 export function discoverProtocol(transport: Service.Transport, options: any, callback: Callback<any>): void;
 export function discoverProtocol(transport: Service.Transport, callback: Callback<any>): void;
 export function extractFileHeader(filePath: string, options?: any): void;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -31,52 +33,121 @@ interface AssembleOptions {
...
+  getSchema(opts: any): any;
+  getTypeName(): any;
+  inspect(): any;
+  isValid(val: any, opts?: any): any;
+  random(): Type;
+  schema(opts?: any): any;
+  toJSON(): any;
+  toString(val?: any): any;
+  wrap(val: any): any;
+  // TODO clone(val, opts)
```
mtth: Can you remove the `TODO`s for methods which are now included?

types/index.d.ts
```diff
@@ -31,52 +33,121 @@ interface AssembleOptions {
...
+  clone(val: any, opts?: any): any;
+  compare(val1: any, val2: any): number;
+  compareBuffers(buf1: any, buf2: any): number;
+  createResolver(type: any, opts?: any): any;
+  decode(buf: any, pos?: any, resolver?: any): any;
+  encode(val: any, buf: any, pos?: any): any;
+  equals(type: any): any;
+  fingerprint(algorithm?: any): any;
+  fromString(str: any): any;
+  getAliases(): any;
```
mtth: All the getters will soon be deprecated (and removed in the next major version), in favor of properties:

+ `getAliases()` -> `aliases`
+ `getName(bool)` -> `name` and `branchName`
+ ...

Can you update the declarations accordingly throughout?

TysonAndre:
https://github.com/mtth/avsc/pull/146 updates package.json, I reopened it.

I see that the wiki has up to date readonly property documentation, I'll update based on that later this week.

types/index.d.ts
TysonAndre: Aside: https://github.com/mtth/avsc/wiki/API#typetypes-1

I don't see a definition of getBranchType() but an example mentions that method. Should that be `constructor.type`?

mtth: Good catch, the example should reference `constructor.type` instead. I'll update the documentation.

types/index.d.ts
```diff
@@ -31,52 +34,100 @@ interface AssembleOptions {
...
+  static forTypes(types: any, opts?: any): Type;
+  static forValue(value: object, opts?: any): Type;
+  static isType(arg: any): boolean;  // TODO remaining args
 }
 
-export namespace Service {
-  interface Protocol {
-    [key: string]: any; // TODO object should be further specified
-  }
+// TODO: Can this project remove Protocol from types completely, it's a deprecated export?
```
mtth: Yes, I think it'd be best to remove the protocol class entirely from here.

types/index.d.ts
```diff
@@ -31,52 +34,100 @@ interface AssembleOptions {
...
+  static isService(any: any): any;
+}
+
+export class Service {
+  constructor(name: any, messages: any, types: any, ptcl: any, server: any);
+  createClient(options?: Partial<Service.ClientOptions>): Service.Client;
+  createEmitter(args: any): any;
+  createListener(args: any): any;
+  createServer(options?: Partial<Service.ServerOptions>): Service.Server;
+  emit(args: any): any;
```
mtth: `emit` and `on` are deprecated and there are a few methods/properties missing (`types`, `type(string)`, etc.). Could update the typings to be in sync with the [API doc](https://github.com/mtth/avsc/wiki/API#class-service), same as you did for the types?

types/index.d.ts
TysonAndre: class Protocol is still referenced by other parts of this file. The references aren't marked as deprecated.

types/index.d.ts
```diff
@@ -126,26 +167,149 @@ export namespace Service {
...
+    random(): LogicalType;
+  }
 
-  function compatible(client: Client, server: Server): boolean;
-  function forProtocol(protocol: Protocol, options: any): Service;
-  function isService(obj: any): boolean;
+  class LongType extends Type {
+    constructor();
+    random(): LongType;
+    __with(methods: object, noUnpack?: boolean;
```
mtth: `static`? Also looks like a closing parens is missing.

types/index.d.ts
```diff
@@ -31,68 +35,112 @@ interface AssembleOptions {
...
+  static forValue(value: object, opts?: any): Type;
+  static isType(arg: any, ...prefix: string[]): boolean;
+  static __reset(size: number): void;
 }
 
-export namespace Service {
-  interface Protocol {
-    [key: string]: any; // TODO object should be further specified
-  }
+// Deprecated, but kept for now because this is referenced elsewhere (e.g. Service.forProtocol())
```
mtth: Actually, all references to `Protocol` in this file should be `Object` (or `any`, not sure how TypeScript treats vanilla objects). For example `readProtocol`'s return value doesn't have any of the methods below. If we wanted to be more accurate, we could define a custom interface with the expected attributes in the object (e.g. `"messages"`, `"protocol"`, and `"types"`) but it's probably fine to do without for now.

types/index.d.ts
```diff
@@ -126,26 +167,149 @@ export namespace Service {
...
+}
+
+export namespace types {
+  class ArrayType extends Type {
+    constructor(schema: Schema, opts: any);
+    readonly itemsType: Type;
+    random(): ArrayType;
+  }
+
+  class BooleanType extends Type {  // TODO: Document this on the wiki
```
mtth: Ah, you're right; these aren't in the Wiki. The reason they were not included originally is they don't have any custom methods (other than those in the `Type` superclass). I'll add them when I get a chance.

PR #146, opened 2018-01-09
Add the `types/` folder to the published NPM package

TysonAndre:
Currently, `tsc --strict` warns for projects with a dependency on
the latest avsc (5.1.2) due to the folder types/ not being included in the
published package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
   ],
   "files": [
     "lib",
-    "etc/browser"
+    "etc/browser",
+    "types"
   ],
   "main": "./lib",
   "types": "./types",


TysonAndre:
Never mind, wasn't aware of https://www.npmjs.com/package/@avro/types

package.json
```diff
@@ -64,5 +66,8 @@
   "repository": {
     "type": "git",
     "url": "git://github.com/mtth/avsc.git"
+  },
+  "dependencies": {
+    "stream": "0.0.2"
```
mtth: What is this for?

TysonAndre:
Most of those changes were intended for another project and accidentally commited while still in this folder.

`@types/node` isn't useful unless a tsconfig.json is added to the project (There's a "checkJs": true option that allows analysis of javascript projects via their JSDoc). Removed that.

mtth:
Thanks!

PR #150, opened 2018-02-13
Fix creating a resolver when a record field name is a javascript keyword

tboothman:
Because the resolver creates cpde that uses field names as variables you get an error if you use a javascript keyword as a record field name when creating a resolver:
```
SyntaxError: Unexpected token void
    at new Function (<anonymous>)
    at RecordType._update (/code/node_modules/avsc/lib/types.js:2385:20)
    at RecordType.Type.createResolver (/code/node_modules/avsc/lib/types.js:523:10)
    at RecordType._update (/code/node_modules/avsc/lib/types.js:2334:30)
    at RecordType.Type.createResolver (/code/node_modules/avsc/lib/types.js:523:10)
```

Fix is a little lazy, but it works and it'd make field names starting with numbers work too. I considered making an object to store the data but that might affect performance.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -2339,7 +2339,7 @@ RecordType.prototype._update = function (resolver, type, opts) {
       } else {
         resolvers[name].push(fieldResolver);
       }
-      innerArgs.push(field.name);
+      innerArgs.push('_' + field.name);
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/types.js
```diff
@@ -2339,7 +2339,7 @@ RecordType.prototype._update = function (resolver, type, opts) {
       } else {
         resolvers[name].push(fieldResolver);
       }
-      innerArgs.push(field.name);
+      innerArgs.push('_' + field.name);
```
mtth: To avoid prefixing the underscore in two different places, can we add it when creating `fieldResolver`'s `name` just above? We can then switch this line to use `fieldResolver.name` and the code below should do the right thing without changes.

tboothman:
There you go. I was pretty scared of changing any of the variables in that method as I couldn't really tell where they ended up.
Tests pass though .. so I guess it's good to go

mtth:
Yes, we should be all good - I'll cut a new release with the fix later today. Thanks again!

Issue #155, opened 2018-02-28
Skipping a logical type field fails when resolving a schema


tboothman:
Decoding a message that was encoded using a schema with a field that's a logical type with a schema that does not have that field fails using the sample code at https://gist.github.com/mtth/1aec40375fbcb077aee7#file-date-js

I fixed it by implementing a _skip method on the logical type's class. Is there any reason why avro.types.LogicalType couldn't implement _skip like this? Would save anyone else running into this problem.

```javascript
_skip(tap) {
    this.getUnderlyingType()._skip(tap);
}
```

Stack trace
```javascript
Error: abstract
    at DateTimeType.abstractFunction (/local/lamp/code/tmp-2018-02-r11/node_modules/avsc/lib/utils.js:210:37)
    at Resolver.readstuff [as _read] (eval at RecordType._update (/local/lamp/code/tmp-2018-02-r11/node_modules/avsc/lib/types.js:2387:20), <anonymous>:8:6)
    at readValue (/local/lamp/code/tmp-2018-02-r11/node_modules/avsc/lib/types.js:2811:21)
    at RecordType.Type.fromBuffer (/local/lamp/code/tmp-2018-02-r11/node_modules/avsc/lib/types.js:575:13)
    at AvroDecoder._transform (/local/lamp/code/tmp-2018-02-r11/transforms/avroDecoder.js:35:41)
```

Message encoded with this schema
```
{
  "name": "stuff",
  "type": "record",
  "fields": [
    {
      "name": "id",
      "type": "long"
    },
    {
      "name": "updateTimeStamp",
      "type": {
        "type": "long",
        "logicalType": "timestamp-millis"
      },
      "default": 0,
    }
]}
```
Message decoded with this schema
```
{
  "name": "stuff",
  "type": "record",
  "fields": [
    {
      "name": "id",
      "type": "long"
    }
]}
```

The logical class I used:
```javascript
class DateTimeType extends avro.types.LogicalType {
    _fromValue(val) {
        return new Date(val);
    }
    _toValue(date) {
        return date instanceof Date ? +date : undefined;
    }
    _resolve(type) {
        if (avro.Type.isType(type, 'long', 'string', 'logical:timestamp-millis')) {
            return this._fromValue;
        }
    }
//    _skip(tap) {
//        this.getUnderlyingType()._skip(tap);
//    }
}
```

mtth:
[...]

PR #156, opened 2018-02-28
Allow logical types to be skipped when resolving schemas that do not …

tboothman:
…have their field. Refs #155

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -2606,6 +2606,10 @@ LogicalType.prototype._deref = function (schema, opts) {
   return schema;
 };
 
+LogicalType.prototype._skip = function(tap) {
+  this.underlyingType._skip(tap);
+};
+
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -241,6 +242,7 @@ export namespace types {
     _resolve(type: Type): any;
     _toValue(any: any): any;
     random(): LogicalType;
+    getUnderlyingType(): Type;
```
mtth: I'm hoping to deprecate this method next major release; can you remove it?

lib/types.js
```diff
@@ -2606,6 +2606,10 @@ LogicalType.prototype._deref = function (schema, opts) {
   return schema;
 };
 
+LogicalType.prototype._skip = function(tap) {
```
mtth: Nit: space between `function` and the first parens for consistency.

PR #162, opened 2018-03-06
Fix line causing an error in typescript 2.7.2

TysonAndre:
Seen in a typescript project using avsc:
(And using --strict)

```
node_modules/avsc/types/index.d.ts(72,3): error TS7010: '_skip', which
lacks return-type annotation, implicitly has an 'any' return type.
```


mtth:
Thanks for the PR @TysonAndre! Released in `5.2.3`.

PR #171, opened 2018-05-01
types: Add encode, use Buffer where possible

alexander-alvarez:
Used the API definitions from https://github.com/mtth/avsc/wiki/API#typeencodeval-buf-pos and https://github.com/mtth/avsc/wiki/API#typedecodebuf-pos-resolver


mtth:
Thanks @alexander-alvarez!

PR #172, opened 2018-05-03
Encoding Namespaced objects with similar names

alexander-alvarez:
Failing test case for an edge case I found, see comments in code for my diagnosis.

https://github.com/mtth/avsc/blob/1c8f7a49abaf91bddc6f8dd3b4f08e2f263cd038/lib/types.js#L2105
gives an unqualified name which causes the function's to override each other.

So in the test case example -- it's trying to encode `record.b` with the encoder for `wrapped.foo` instead of the encoder from `foo`

/cc @mtth

### DIFF
--- a/test/test_types.js
+++ b/test/test_types.js
@@ -3556,6 +3556,41 @@ suite('types', function () {
       assert.throws(function () { t.encode('hi', buf, -1); });
     });
 
+    ['fooWrapped', 'wrapped.foo'].forEach(function(name) {
+
+      test('nested schema with registry: ' + name, function() {
+        var a = {
+          name: name,
+          type: 'record',
+          fields: [
+            { name: 'a', type: 'boolean'},
+            { name: 'b', type: 'foo'}
+          ]
+        }
+        
+        var b = {
+          name: 'foo',
+          type: 'record',
+          fields: [
+            { name: 'c', type: 'boolean'}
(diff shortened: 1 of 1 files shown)


test/test_types.js
```diff
@@ -3556,6 +3556,41 @@ suite('types', function () {
...
 
+    ['fooWrapped', 'wrapped.foo'].forEach(function(name) {
+
+      test('nested schema with registry: ' + name, function() {
+        var a = {
+          name: name,
+          type: 'record',
+          fields: [
+            { name: 'a', type: 'boolean'},
+            { name: 'b', type: 'foo'}
```
mtth: The issue here is that the type `foo` is referenced unqualified, so according to [Avro resolution rules](https://avro.apache.org/docs/1.8.2/spec.html#names) it inherits the namespace of the outer type. When you use `wrapped.foo` as name for the outer type, this field therefore becomes namespaced under `wrapped` as well (i.e. becomes `wrapped.foo`) and ends up referencing the outer type.

One way to achieve what you're trying to do is to prefix the reference with a dot (`.foo`). In that case it will always refer to the registry's `foo` entry.

alexander-alvarez: @mtth  Thanks for the clarification, I learned something new today. Thanks for the awesome lib.

PR #173, opened 2018-05-07
types: Declaring methods for UnionTypes

alexander-alvarez:
Including the methods defined here

https://github.com/mtth/avsc/blob/master/lib/types.js#L2539

and

https://github.com/mtth/avsc/blob/master/lib/types.js#L1198


mtth:
Thank you!

PR #175, opened 2018-05-11
Update of typescript definitions

reuzel:
Hi,

I took the liberty to update the typescript definitions. They were not working for me, so I corrected them as far as I could. 

I tried to use generics on the types, but reverted those changes for now as these become awkward with the field definitions in the types (not sure how to create an array of generic field types, where the generic parameter is not set yet...).

There may be more TODOs but, I think this one is an improvement...

Regards,
Joost

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -157,9 +157,9 @@ export class Type {
   compare(val1: any, val2: any): number;
   compareBuffers(buf1: Buffer, buf2: Buffer): number;
   createResolver(type: Type): Resolver;
-  decode(buf: Buffer, pos?: number, resolver?: Resolver): any;
+  decode(buf: Buffer, pos?: number, resolver?: Resolver): { value: any, offset: number};
   encode(val: any, buf: Buffer, pos?: number): void;
-  equals(type: any): boolean;
+  equals(type: Type): boolean;
   fingerprint(algorithm?: string): Buffer;
   fromBuffer(buffer: Buffer, resolver?: Resolver, noCheck?: boolean): any;
   fromString(str: string): any;
@@ -168,7 +168,7 @@ export class Type {
   random(): Type;
   schema(opts?: Partial<SchemaOptions>): Schema;
   toBuffer(value: any): Buffer;
-  toJSON(): string;
+  toJSON(): object;
   toString(val?: any): string;
   wrap(val: any): any;
   readonly aliases: string[] | undefined;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -4,43 +4,40 @@
 // TODO: Wherever the argument names are just `args: any`, it was probably generated from the signature of util.deprecate. Fix argument counts and types.
 // NOTE: This does not contain entries for functions available in the browser (functions/methods in etc/browser)
 
-import * as stream from 'stream';
+import * as stream from 'stream'; 
 import { EventEmitter } from 'events'
 
 //"virtual" namespace (no JS, just types) for Avro Schema
-declare namespace schema
-{
+declare namespace schema {
   export type AvroSchema = DefinedType | DefinedType[];
-  type DefinedType = PrimitiveType | ComplexType | LogicalType; 
+  type DefinedType = PrimitiveType | ComplexType | LogicalType | string;
   type PrimitiveType = 'null' | 'boolean' | 'int' | 'long' | 'float' | 'double' | 'bytes' | 'string';
   type ComplexType = NamedType | RecordType | EnumType | MapType | ArrayType | FixedType;
   type LogicalType = ComplexType & LogicalTypeExtension;
-  
-  interface NamedType
-  {
+
+  interface NamedType {
     type: PrimitiveType
   }
-  
-  interface RecordType
-  {
+
+  interface RecordType {
     type: "record";
     name: string;
-    namespace? : string;
+    namespace?: string;
     doc?: string;
     aliases?: string[];
     fields: {
-          name: string;
-          doc?:string;
-          type: AvroSchema;
-          default?: any;
-        }[];
-    order? : "ascending" | "descending" | "ignore";
+      name: string;
+      doc?: string;
+      type: AvroSchema;
+      default?: any;
+    }[];
+    order?: "ascending" | "descending" | "ignore";
   }
 
   interface EnumType {
     type: "enum";
     name: string;
-    namespace?:string;
+    namespace?: string;
     aliases?: string[];
     doc?: string;
     symbols: string[];
@@ -62,10 +59,9 @@ declare namespace schema
     aliases?: string[];
     size: number;
   }
-  interface LogicalTypeExtension
-  {
+  interface LogicalTypeExtension {
     logicalType: string;
-    [param:string] : any;
+    [param: string]: any;
   }
 }
 //Types of Options/arguments
@@ -75,17 +71,17 @@ type Schema = Type | schema.AvroSchema;
 type Callback<V, Err = any> = (err: Err, value: V) => void;
 
 
-type Codec = (buffer: Buffer, callback: Callback<Buffer>) => void; 
+type Codec = (buffer: Buffer, callback: Callback<Buffer>) => void;
 
 interface CodecOptions {
-  [name:string] : Codec;
+  [name: string]: Codec;
 }
 
 interface DecoderOptions {
   noDecode: boolean;
   readerSchema: string | object | Type;
   codecs: CodecOptions;
-  parseHook: (schema:Schema) => Type
+  parseHook: (schema: Schema) => Type
 }
 
 interface EncoderOptions {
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -7,80 +7,188 @@
 import * as stream from 'stream';
 import { EventEmitter } from 'events'
 
-type Schema = string | object;  // TODO object should be further specified
+//"virtual" namespace (no JS, just types) for Avro Schema
+declare namespace schema
+{
```
mtth: Can you put the opening brace on the same line as the declaration for consistency with the rest of the file? (Throughout too.)

types/index.d.ts
```diff
@@ -7,80 +7,188 @@
...
+  {
+    type: "record";
+    name: string;
+    namespace? : string;
+    doc?: string;
+    aliases?: string[];
+    fields: {
+          name: string;
+          doc?:string;
+          type: AvroSchema;
```
mtth: All nested `AvroSchema`s could be relaxed to `Schema` (`Type`s are also accepted).

types/index.d.ts
```diff
@@ -7,80 +7,188 @@
 import * as stream from 'stream';
 import { EventEmitter } from 'events'
 
-type Schema = string | object;  // TODO object should be further specified
+//"virtual" namespace (no JS, just types) for Avro Schema
+declare namespace schema
+{
+  export type AvroSchema = DefinedType | DefinedType[];
```
mtth: Does this allow references? (References can be arbitrary strings, pointing to values in a registry.)

types/index.d.ts
reuzel: ah, styling. Will do.

types/index.d.ts
reuzel: Schema lives in a namespace outside this one (one level up).  Would that not violate layering? I thought to make this namespace completely independent on how it is being used. It is just a typescript representation of the Avro spec. Or am I misunderstanding you?

types/index.d.ts
reuzel: ooh, perhaps not. I haven't used references in my project so far. It would be easy to fix, to add a string type to the DefinedType (along with the primitive, complex and logical types). The primitive type than becomes a kind of redundant as they are just a collection of strings, though it still helps in code-completion scenarios.

jsphweid:
Also, should support for `.on('finish', () => etc.)` be added?

types/index.d.ts
mtth: It's sometimes convenient to nest types inside of a schema declaration (a simpler alternative to passing them through `Type.forSchema`'s registry option). I'm not sure whether the namespacing is worth the extra restriction; it might also break existing code.

types/index.d.ts
```diff
@@ -172,22 +271,81 @@ export namespace Service {
...
 export namespace streams {
-  class BlockDecoder {
-    constructor(opts?: any);
-    static defaultCodecs(): any;
+
+  class BlockDecoder extends stream.Duplex {
+    constructor(opts?: Partial<DecoderOptions>);
+    static defaultCodecs(): CodecOptions;
+
+    //adding all listeners, etc. Can't add a single method override for metadata, all base methods need to be repeated :-(
```
mtth: Are these really required? How come they aren't provided by `Duplex`?

types/index.d.ts
```diff
@@ -4,83 +4,182 @@
...
-  constructor(schema: Schema, opts?: any);
-  createResolver(type: any, opts?: any): any;  // TODO: opts not documented on wiki
-  decode(buf: Buffer, pos?: number, resolver?: any): any;
-  encode(val: any, buf: Buffer, pos?: number): void; 
-  fingerprint(algorithm?: any): any;
-  fromBuffer(buffer: Buffer, resolver?: any, noCheck?: boolean): Type; // TODO
-  fromString(str: any): any;
+  createResolver(type: Type): Resolver;
+  decode(buf: Buffer, pos?: number, resolver?: Resolver): { value: any, offset: number};
+  encode(val: any, buf: Buffer, pos?: number): void;
```
mtth: `encode` returns a `number` (the new offset in the buffer).

types/index.d.ts
reuzel: The implementation of most of these is provided by Duplex. However, in the definition files, Typescipt requires you to specify the overload function (the one with string, (...args:any[]) ). Once you do that, all the events from the base-class (Duplex) are hidden from code completion (at least in vscode). To make them visible again, you have to add them all :-(

I see this as a Typescript or otherwise vscode bug...

types/index.d.ts
reuzel: Will change.

types/index.d.ts
mtth: That's a bummer. It might be surprising to add them here but not to the other duplex streams below; I'd prefer staying consistent and not adding them anywhere, at the cost of losing auto-completion (until the issue you mention is fixed) - do you mind you take them out? (Adding them everywhere would be an unreasonable amount of clutter for the benefits.)

alexander-alvarez:
what's the status here? would like to see the types updates wrapped up (or reverted) and released if possible

reuzel:
Sorry,  but I’ve been extremely busy lately. I will come back to this as soon as I can. Expect the changes to be made within a few days from now.

Regards,
Joost

Op 1 jun. 2018 om 20:29 heeft Alex Alvarez <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


what's the status here? would like to see the types updates wrapped up (or reverted) and released if possible

reuzel:
Last question, before making the changes: The BlockDecoder is the only duplex stream where the addition of all the events is necessary, as it is the only one that adds an event. The repetition of events is only necessary when an extension is made. The event override 'hides' the events from the super classes (as far as those exist in JavaScript).  As the BlockDecoder is the only one that emits a new type of event, it is the only one for which the repetition is required. Given that, there are two choices:

1. Remove all the listeners: the metadata event will work type-check wise but is not visible in any code-completion. The Duplex events will be visible.
2. Add the metadata event, hide all the standard Duplex events.
3. Add the meta-data event and repeat the definition of all the standard duplex events (as in the current PR).

None of them are as elegant as  you would wish, but I leave the choice to you...

Regards,
Joost

mtth:
Thank you for the detailed explanation @reuzel, this makes sense. Can we go with the first option, adding a comment linking to this PR mentioning why the `metadata` event is not declared? It's unfortunate that code-completion will not display the event, but the benefits outweigh the inconvenience IMHO.

reuzel:
@mtth The PR should now be in a state that it can  be merged.

alexander-alvarez:
@reuzel  worth noting you can do a repository pattern like so https://github.com/Blizzard/node-rdkafka/pull/420/files#diff-b52768974e6bc0faccb7d4b75b162c99R21
so at least the clutter moves somewhere else?

/shrug

mtth:
Merged; thank you @reuzel!

types/index.d.ts
```diff
@@ -4,95 +4,194 @@
...
+
 export function assembleProtocol(filePath: string, opts: Partial<AssembleOptions>, callback: Callback<object>): void;
 export function assembleProtocol(filePath: string, callback: Callback<object>): void;
-export function combine(args: any): any;
-export function createFileDecoder(fileName: string, codecs?: Partial<CodecOptions>): Decoder;
-export function createFileEncoder(filePath: string, schema: Schema, options?: any): Encoder;
+export function createFileDecoder(fileName: string, opts?: Partial<DecoderOptions>): streams.BlockDecoder;
+export function createFileEncoder(filePath: string, schema: Schema, opts?: Partial<EncoderOptions>): streams.BlockEncoder;
+export function createBlobEncoder(schema: Schema, opts?: Partial<EncoderOptions>): stream.Duplex;
+export function createBlobDecoder(blob: Blob, opts?: Partial<DecoderOptions>): streams.BlockDecoder;
```
alexander-alvarez: FWIW I had to add `dom` as a `lib` to my `tsconfig.json` for typescript to compile O.K. after this change.
Not sure if this is something we can work around though in this file. @reuzel 

    "lib": ["dom"]

alexander-alvarez: I think a triple slash directive should make it work, but it's not taking effect on my local
    `/// <reference types="dom" />`

reuzel: The triple slash syntax is rather outdated. I never use it anymore. In this case, the lib option seems the right way forward. It is just telling the typescript compiler that the dom objects are assumed to be available. Without, the Blob type would be unknown.

Why do you see this as an issue? As far as I know, the lib option is not transitive, in the sense that clients of your library do not need to add them as long as they do not include the Blob encoders/decoders. As you have nicely splitted that functionality into a browser lib, I don't really see an issue?

By the way, I didn't encounter this issue in my own project. I'm using webpack for the client pieces, and when set in the right mode, WebPack enables the dom objects automatically.

PR #177, opened 2018-05-15
Update index.d.ts to accept 'finish' flag

brianfitzgerald:


types/index.d.ts
```diff
@@ -21,6 +21,7 @@ export interface CodecOptions {
 export interface Decoder {
   on(type: 'metadata', callback: (type: Type) => void): this;
   on(type: 'data', callback: (value: object) => void): this;
+  on(type: 'finish', callback: (value: object) => void): this;
```
reuzel: Out of curiosity: Why would you want to listen to the finish event when decoding? Would it not be more logical to listen to the 'end' event? I added that one in my pull request already. If 'finished' is interesting too, we could merge the pull-requests into one...

brianfitzgerald: Good call, I can add the end as well. I'm doing this PR since this is causing errors we're having to suppress on production code, and your PR looks like it's pretty comprehensive, so I figured this one would be easier to merge in first.

reuzel: Both finished and end are supported by BlockDecoder as it is a Duplex stream. I was just wondering if I should add the Writable events to the BlockDecoder definitions. I noticed the Writable events are not explicitly added to the official Duplex definitions either, that's why I skipped them.

mtth:
Merging this PR first since it's tiny - thanks @brianfitzgerald!

PR #178, opened 2018-05-15
types: More type defintions

alexander-alvarez:
`toJSON` was actually wrong, which is worse than too generic types (i.e. any).

`Type.equals` method wasn't defined in the typings file

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -69,7 +69,7 @@ export class Type {
   schema(opts?: any): object;
   toBuffer(value: object): Buffer;
   toJSON(): object;
-  toString(val?: any): any;
+  toString(val?: any): string;
   wrap(val: any): any;
   _skip(tap: any): any;
   readonly aliases: string[]|undefined;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,18 +57,18 @@ export class Type {
   compareBuffers(buf1: Buffer, buf2: Buffer): number;
   constructor(schema: Schema, opts?: any);
   createResolver(type: any, opts?: any): any;  // TODO: opts not documented on wiki
-  decode(buf: Buffer, pos?: number, resolver?: any): any;
+  decode(buf: Buffer, pos?: number, resolver?: any): { value: any, offset: number};
   encode(val: any, buf: Buffer, pos?: number): void;
   equals(type: Type): boolean;
   fingerprint(algorithm?: string): Buffer | string;
   fromBuffer(buffer: Buffer, resolver?: any, noCheck?: boolean): Type; // TODO
   fromString(str: any): any;
   inspect(): string;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -57,18 +57,19 @@ export class Type {
   compareBuffers(buf1: Buffer, buf2: Buffer): number;
   constructor(schema: Schema, opts?: any);
   createResolver(type: any, opts?: any): any;  // TODO: opts not documented on wiki
-  decode(buf: Buffer, pos?: number, resolver?: any): any;
-  encode(val: any, buf: Buffer, pos?: number): void; 
-  fingerprint(algorithm?: any): any;
+  decode(buf: Buffer, pos?: number, resolver?: any): { value: any, offset: number};
+  encode(val: any, buf: Buffer, pos?: number): void;
+  equals(type: Type): boolean;
+  fingerprint(algorithm?: string): Buffer | string;
```
reuzel: As far as I can see fingerprint can return Buffers only

types/index.d.ts
```diff
@@ -57,18 +57,19 @@ export class Type {
   compareBuffers(buf1: Buffer, buf2: Buffer): number;
   constructor(schema: Schema, opts?: any);
   createResolver(type: any, opts?: any): any;  // TODO: opts not documented on wiki
-  decode(buf: Buffer, pos?: number, resolver?: any): any;
-  encode(val: any, buf: Buffer, pos?: number): void; 
-  fingerprint(algorithm?: any): any;
+  decode(buf: Buffer, pos?: number, resolver?: any): { value: any, offset: number};
```
reuzel: Decode: Good catch, I added that to my pullrequest as well, hoping all of this gets committed faster to the project.

types/index.d.ts
alexander-alvarez: I think so too. I was deferring to https://nodejs.org/api/crypto.html#crypto_class_hash => https://nodejs.org/api/stream.html#stream_readable_read_size
but more than happy to change

types/index.d.ts
mtth: I'm OK either way. If @reuzel doesn't mind updating #175, I can merge this once `fingerprint`'s signature is changed; let me know what you prefer.

types/index.d.ts
mtth: Let's change it to just `Buffer`.

types/index.d.ts
reuzel: done

mtth:
@alexander-alvarez - are there still changes from this PR you wanted to merge or is this covered by https://github.com/mtth/avsc/pull/175?

alexander-alvarez:
nope they're incorporated :+1:

PR #193, opened 2018-07-16
Align with apache commons avro implementation: Don't validate namespace when qualifying name

kvsikand:
Upon investigating the way name validation works in AVSC vs. the Apache Commons Java implementation of the Avro spec, I noticed a discrepancy.
Specifically, when qualifying a name/namespace, avsc validates **both** the _name_ and the _namespace_. This is because the _qualify_ method adds the namespace to the name **before** doing any validation.
The Java implementation in Apache Commons, however, behaves differently. As seen here: [Java](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L512-L518), this standard implementation only validates the _name_, not the _namespace_ portion. This means that when the name/namespace combination looks something like 'test_name','test.1.namespace', avsc will consider this an error case, whereas the Apache Commons implementation will not.

This pull request seeks to resolve this discrepancy, and add appropriate tests.

The Avro spec isn't super specific about this case, but on first reading it feels like AVSC is actually **more** strictly following the spec than the Apache commons implementation, but seeing as the latter is more "standard", it feels like making the change here to align with that established library is more reasonable.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -2876,16 +2876,20 @@ function unqualify(name) {
  * @param namespace {String} Optional namespace.
  */
 function qualify(name, namespace) {
-  if (~name.indexOf('.')) {
+  var globalNamespace = ~name.indexOf('.');
+  if (globalNamespace) {
     name = name.replace(/^\./, ''); // Allow absolute referencing.
-  } else if (namespace) {
-    name = namespace + '.' + name;
   }
-  name.split('.').forEach(function (part) {
+  name.split('.').forEach(function(part) {
     if (!isValidName(part)) {
       throw new Error(f('invalid name: %j', name));
     }
   });
+
+  if (!globalNamespace && namespace) {
+    name = namespace + '.' + name;
+  }
+
   var tail = unqualify(name);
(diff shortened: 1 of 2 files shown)


mtth:
Hi @kvsikand, thank you for the PR and detailed explanation. I don't think this is a good change unfortunately; the spec is actually pretty clear and--as you point out--matches `avsc`'s behavior (from the current [`1.8.2`](https://avro.apache.org/docs/1.8.2/spec.html#names), emphasis mine):

> The name portion of a fullname, record field names, and enum symbols must:
>
> + start with [A-Za-z_]
> + subsequently contain only [A-Za-z0-9_]
>
> A namespace is a dot-separated sequence __of such names__.

Unless the specification changes, a PR to the Apache Commons implementations seems more appropriate. This change would also have the drawback of breaking the equivalence of (namespace, unqualified name) and (qualified name) combinations: `{namespace: '1.one', name: 'Foo'}` would be valid but `{name: '1.one.Foo'}` would not.

Issue #212, opened 2018-11-13
Type.schema() doesn't return defaultValue


davesalomon:
Thanks for the great work on this repo.

I've noticed that a default value isn't coming back when I call `.schema()` off a type. In the example below, I would expect `{ name: 'someValue', type: 'int', default: 0 }`.

I _think_ I've created a fix for this, and will create a PR shortly.

```
const avro = require('avsc');

const avroSchema = avro.Type.forSchema({
  type: 'record',
  name: 'my_test_topic',
  fields: [
    { "name": "testRecordId", "type": "string" },
    { "name": "someValue", "type": "int", default: 0 },
    {
      "name": "metadata",
      "type": {
        "type": "record",
        "name": "EventMetadata",
        "fields": [
          {
            "name": "eventId",
            "type": "string"
          },
          {
            "name": "traceToken",
            "type": "string"
          },
          {
            "name": "createdAt",
            "type": {
              "type": "long",
              "logicalType": "timestamp-millis"
            }
          }
        ]
      }
    }
  ]
});

console.log(avroSchema.schema());
// { name: 'my_test_topic',
//   type: 'record',
//   fields: 
//    [ { name: 'testRecordId', type: 'string' },
//      { name: 'someValue', type: 'int' },   <--- Missing   default: 0
//      { name: 'metadata', type: [Object] } ] }

console.log(avroSchema.fields[1].defaultValue());
// 0
```

davesalomon:
When fixing the tests, I noted that one of had `// Stripped defaults.`.

I don't really see why this would be desired behaviour, but if I'm wrong, please do correct me...

mtth:
Hey @davesalomon. By default [`type.schema`](https://github.com/mtth/avsc/wiki/API#typeschemaopts) returns the type's [canonical schema](https://avro.apache.org/docs/current/spec.html#Parsing+Canonical+Form+for+Schemas), which the specification defines as excluding defaults (see the "STRIP" transformation). I believe you can get the behavior you want by setting the `exportAttrs` option:

```js
type.schema(); // Canonical schema, without defaults.
type.schema({exportAttrs: true}); // Expanded schema, including defaults (and more).
```

davesalomon:
That's my bad then - many thanks, I'll close this issue and ditch the PR!

PR #213, opened 2018-11-13
Add defaultValue to returned JSON from .schema() call

davesalomon:
Fixes https://github.com/mtth/avsc/issues/212


davesalomon:
Not actually an issue.

PR #224, opened 2019-02-25
refactored lib/types and lib/utils to make rollup friendly

hath995:
This is more of a request for comment than a real pull request. I would like to use avsc in the browser using the packager Rollup. Currently, I could not get that to work to with the standard distribution. I did try browserify and that worked of course, but it doesn't really help me with my bundle.

I'm also trying to minimize the weight of the library. With the included build.js and rollup I could push the size down to 68k after this refactor (which is mostly neutral but may have broken one or two things)

So in this pr I removed as much node.js code as I could from lib/util and lib.types. To do that I refactored the types to the es6 class syntax to remove the dependency on util.inherits. As removing the node.js formatting and debug functions. There are two places which I added @FIXME notices where that probably broke something. I haven't run any of the tests yet.

Overall, it would be helpful if there was an esm distribution of avsc, which would probably be a different way to make this library rollup friendly. 

So what would you recommend on how to proceed?

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,0 +1,29 @@
+const rollup = require('rollup');
+const resolve = require('rollup-plugin-node-resolve');
+const cjs = require('rollup-plugin-commonjs');
+const terser = require('rollup-plugin-terser').terser;
+const builtins = require('rollup-plugin-node-builtins');
+
+async function build() {
+  const avsc = await rollup.rollup({
+    input: './etc/browser/avsc-types.js',
+    plugins: [
+      builtins(),
+      //nodeglobals(),
+      resolve({browser: true /*,preferBuiltins: false*/ }),
+      cjs(),
+      terser(),
+    ]
+  })
+    
+  await avsc.write({
+    name: "leader",
+    format: 'iife',
+    file: './avsc.js',
+    sourcemap: true
+  });
+
+}
+build().catch(err => {
+  console.error(err)
+})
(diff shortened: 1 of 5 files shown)


mtth:
Hi @hath995, TIL about `rollup`; thanks for the PR.

Is it possible to split the "weight reduction" part of this PR into a separate PR? Supporting `rollup` sounds like a net win but I am hesitant to also introduce an ES6 requirement in `avsc`.

If you only care about the de/serialization part of this library, another option would be to merge this into [`@avro/types`](https://www.npmjs.com/package/@avro/types) instead of `avsc`. Currently it is identical to the types portion of `avsc` (see the [`packages`](https://github.com/mtth/avsc/tree/packages) branch in this repository) but going forward I am planning on adding a few features, probably relying on ES6.

hath995:
I can do a more minimal pr. 

I think moving to es6 is generally a good idea. Given how principled this library is about types it should also be pretty easy to convert it to Typescript as well. Then folks could compile it to whatever target language version they wanted. Babel could do the same though.

One issue I can describe in more detail now. Here
 https://github.com/hath995/avsc/blob/rollup/lib/types.js#L2694
You initialize a value on `this` before you call the parent class constructor which breaks in es6 which disallows assigning values to `this` before calling the parent constructor. As a way of future proofing the library it would be good to find an alternative there.

PR #226, opened 2019-03-13
isSafeLong now accepts int64 length

jeffbeagley:
When topics have dates encoded as Int64 and are nanoseconds since Epoch, the length is then 19 digits long exceeding the current min/max as defined in isSafeLong. Changed isSafeLong to return true when long is within range outlined here, https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/integral-types-table

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -2984,7 +2984,7 @@ function readArraySize(tap) {
  * + We must remove one from each bound because of rounding errors.
  */
 function isSafeLong(n) {
-  return n >= -9007199254740990 && n <= 9007199254740990;
+  return n >= -9223372036854775808 && n <= 9223372036854775807;
 }
 
 /**


mtth:
Hi @jeffbeagley, this isn't a safe change unfortunately. JavaScript runtimes will silently convert these numbers to doubles and likely cause precision loss. For example:

```javascript
9223372036854775806 === 9223372036854775807 // True!
```

You can support nanosecond-resolution timestamps by using a custom [`LongType`](https://github.com/mtth/avsc/wiki/API#class-longtypeschema-opts). The [wiki](https://github.com/mtth/avsc/wiki/Advanced-usage#custom-long-types) has a few examples which should help you get started.

jeffbeagley:
Thanks for reviewing. I had checked out the custom types however couldn't determine how to use that just for decoding purposes.

PR #260, opened 2019-10-06
Add type defs for TransportFunction

luma:
These were `TODO` so I had a go at adding them as it
was casuing some ugliness when I used avsc from Typescript.

These seem to be correct, and definitely work for me, but it
would be great if someone more familar with the codebase
could check my work :-)

It would have been nice to also constrain the type of
streams further depending on whether the transport would
be read, write, or duplex. I had a quick look at this but
it turned out to be a larger change so I decided to back
off and go with a smaller diff.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -238,7 +238,10 @@ export namespace Service {
     objectMode: boolean;
   }
 
-  type TransportFunction = () => void; // TODO
+  
+
+  type TransportFunctionCallback = (err: Error|null|undefined, res?: stream.Stream) => void;
+  type TransportFunction = (cb: TransportFunctionCallback) => stream.Stream; // TODO
 
   type Transport = stream.Duplex | TransportFunction;
 


types/index.d.ts
```diff
@@ -238,7 +238,10 @@ export namespace Service {
     objectMode: boolean;
   }
 
-  type TransportFunction = () => void; // TODO
+  
+
+  type TransportFunctionCallback = (err: Error|null|undefined, res?: stream.Stream) => stream.Stream;
+  type TransportFunction = (cb: TransportFunctionCallback) => void; // TODO
```
mtth: I think the return types are swapped. The outer function is the one which is expected to return a stream:

```typescript
type TransportFunctionCallback = (err: Error|null|undefined, res?: stream.Stream) => void;
type TransportFunction = (cb: TransportFunctionCallback) => stream.Stream;
```

luma: Doh! You're right, that makes way more sense. I've pushed a fix.

mtth:
Merged; thanks @luma.

PR #261, opened 2019-10-08
Allow idl files to recognize default values on enums

rdelvallej32:
I realized that the package does not support Avro 1.9.0 Specification yet, but I wanted to open up this PR to allow the parsing of files to not error at the very least when it sees default values for enums. https://avro.apache.org/docs/current/spec.html#Enums.

At a minimum, this will allow the parsing of a file to continue without a failure because of an unrecognized token. If you wish to add the ability to recognize and use the default value, it will take a little bit more time on my end.

### DIFF
--- a/lib/specs.js
+++ b/lib/specs.js
@@ -478,6 +478,10 @@ Reader.prototype._readEnum = function (schema) {
   do {
     schema.symbols.push(tk.next().val);
   } while (!tk.next({val: '}', silent: true}) && tk.next({val: ','}));
+
+  if (tk.next({val: '=', silent: true})) {
+    schema.default = tk.next().val;
+  }
   return schema;
 };
 
--- a/lib/specs.js
+++ b/lib/specs.js
@@ -481,6 +481,7 @@ Reader.prototype._readEnum = function (schema) {
 
   if (tk.next({val: '=', silent: true})) {
     schema.default = tk.next().val;
+    tk.next({val: ';'});
   }
   return schema;
 };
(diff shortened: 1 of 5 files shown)


lib/specs.js
```diff
@@ -478,6 +478,10 @@ Reader.prototype._readEnum = function (schema) {
   do {
     schema.symbols.push(tk.next().val);
   } while (!tk.next({val: '}', silent: true}) && tk.next({val: ','}));
+
+  if (tk.next({val: '=', silent: true})) {
+    schema.default = tk.next().val;
+  }
```
mtth: It looks like the spec requires a semicolon in the case with a default value, we should add a statement to parse it (`tk.next({val: ';'});`).

test/dat/Hello.avdl
```diff
@@ -16,7 +16,7 @@ protocol Simple {
     FOO,
     BAR, // the bar enum value
     BAZ
-  }
+  } = BAZ
```
mtth: Rather than replace this enum, can you add a second one with a default? This will guard better against regressions.

rdelvallej32: @mtth , all set! I added a new Enum with the default value case

lib/specs.js
rdelvallej32: I added the statement to parse the semicolon when it finds an enum with a default value. Can you point to me in the avro spec where it mentions it? I wasn't able to find it for my reference. Thanks!

arminhaghi: https://avro.apache.org/docs/1.9.1/idl.html#format_enums

arminhaghi:
@mtth if you have no concerns would you mind merging this?

mtth:
Thanks @rdelvallej32. Merged and released in `5.4.14`.

PR #262, opened 2019-10-08
Adding clarification with respect to avro-js

blackdynamo:
Hey!

I created a PR so you don't have to do the work, although it's not much! Hope the wording works and if merged, hope it helps future people.


mtth:
Hey @blackdynamo. Thanks for sending this out. I don't think the top of the README is the best place for the note though, this hasn't been much of an issue in the past couple years. I'll need to think about the best place to add it, perhaps in the Wiki.

PR #264, opened 2019-10-10
Feature/enable enum default values

rdelvallej32:
- Adds test for validating a default value applies when an EnumType's schema has one
- Adds default value check for EnumType

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-node_modules
--- a/lib/types.js
+++ b/lib/types.js
@@ -1669,7 +1669,7 @@ EnumType.prototype._read = function (tap) {
 EnumType.prototype._skip = function (tap) { tap.skipLong(); };
 
 EnumType.prototype._write = function (tap, val) {
-  const index = this._indices[val];
+  var index = this._indices[val];
   var value = index !== undefined
     ? index
     : this._indices[this.defaultValue];
--- a/lib/types.js
+++ b/lib/types.js
@@ -1687,8 +1687,8 @@ EnumType.prototype.compare = function (val1, val2) {
 
 EnumType.prototype._update = function (resolver, type) {
   if (
-    type.getTypeName() !== 'enum' ||
-    (type._name || ~getAliases(this).indexOf(type._name))
+    type.getTypeName() !== 'enum' &&
+    (!type.name || ~getAliases(this).indexOf(type.name))
   ) {
     return;
   }
--- a/lib/types.js
+++ b/lib/types.js
@@ -1629,7 +1629,6 @@ function EnumType(schema, opts) {
 
   this.symbols = Object.freeze(schema.symbols.slice());
   this._indices = {};
-  this.defaultValue = undefined;
   this.symbols.forEach(function (symbol, i) {
     if (!isValidName(symbol)) {
       throw new Error(f('invalid %s symbol: %j', this, symbol));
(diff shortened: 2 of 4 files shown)


rdelvallej32:
Hi @mtth , I took a stab at applying the logic for EnumTypes to handle default values. If there's something that I am missing or if this is incorrect. Please let me know!

lib/types.js
```diff
@@ -1626,15 +1626,22 @@ function EnumType(schema, opts) {
   if (!Array.isArray(schema.symbols) || !schema.symbols.length) {
     throw new Error(f('invalid enum symbols: %j', schema.symbols));
   }
+
   this.symbols = Object.freeze(schema.symbols.slice());
   this._indices = {};
+  this.defaultValue = undefined;
```
mtth: Can you switch to `defaultSymbol`? It would be slightly more specific and consistent.

lib/types.js
```diff
@@ -1626,15 +1626,22 @@ function EnumType(schema, opts) {
...
   this.symbols.forEach(function (symbol, i) {
     if (!isValidName(symbol)) {
       throw new Error(f('invalid %s symbol: %j', this, symbol));
     }
     if (this._indices[symbol] !== undefined) {
       throw new Error(f('duplicate %s symbol: %j', this, symbol));
     }
+
+    if (schema.default && symbol === schema.default) {
```
mtth: Rather than check here for each symbol, let's assign once just before `Object.freeze` below:

```js
if (schema.default !== undefined && this._indices[schema.default] === undefined) {
  throw new Error(f('invalid default symbol: %j', schema.default));
}
this.defaultSymbol = schema.default;
```

Validating once at type construction time also helps by saving us from having to do it repeatedly later when reading or writing.

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
mtth: `node_modules` is better ignored from a more global location. Can you remove this file?

lib/types.js
```diff
@@ -1663,10 +1670,13 @@ EnumType.prototype._skip = function (tap) { tap.skipLong(); };
 
 EnumType.prototype._write = function (tap, val) {
   var index = this._indices[val];
-  if (index === undefined) {
```
mtth: If I understand the spec correctly, default enum symbols are only used during [schema resolution](http://avro.apache.org/docs/current/spec.html#Schema+Resolution) when a writer's symbol does not exist in the reader. We'll therefore need to modify `_update` rather than `_write`.

Instead of the `every` check that makes sure each writer symbol is in the reader, we could iterate over the writer's symbols and insert the reader's default if available when the symbol isn't found in the reader (rather than failing). Something like:

```js
EnumType.prototype._update = function (resolver, type) {
  if (
    type.typeName !== 'enum' ||
    (!type.name || ~getAliases(this).indexOf(type.name))
  ) {
    return;
  }
  var symbols = type.symbols.slice();
  var i, l, symbol;
  for (i = 0, l = symbols.length; i < l; i++) {
    symbol = symbols[i];
    if (this._indices[symbol] !== undefined) {
      continue; // Both the reader and the writer have this symbol.
    }
    if (this.defaultSymbol === undefined) {
      return; // Incompatible enums.
    }
    symbols[i] = this.defaultSymbol; // Substitute the default.
  }
  resolver.symbols = symbols;
  resolver._read = type._read; 
};
```

lib/types.js
rdelvallej32: All Set

.gitignore
rdelvallej32: Okay

lib/types.js
rdelvallej32: Thank you for the feedback, I see what you mean now. It should be reflected in the PR now

mtth:
Superseded by https://github.com/mtth/avsc/commit/e16dbf642c5999c7e5ab09c536d7310f6b7a1fde.

Issue #118, opened 2017-07-16
JSON representation with "left-out" optional fields fails Avro Schema


kdallmeyer-sr:
I don't know if this is an intentional feature or not but I think a usability improvement to the library would be to support JSON objects that leave out optional fields.  Here is an example.

Schema:
```js
var schema = {
  "type" : "record",
  "name" : "Demo",
  "namespace" : "com.shoprunner.data.demo",
  "fields" : [ {
    "name" : "name",
    "type" : "string"
  }, {
    "name" : "count",
    "type" : [ "null", "int" ]
  }, {
    "name" : "price",
    "type" : [ "null", "float" ]
  } ]
};
```
Then I construct a minified version of that object in Javascript:
```js
var data = { name: "Ken" };
```
I would like to represent that data in the Demo schema but it fails validation.
```js
var buf = schema.toBuffer(data);
// Error: invalid ["null","int"]: undefined

var isValid = schema.isValid(data);
// Return: false
```
Using a full object works.  But setting the fields with null, seems like work that probably doesn't need to be necessary.
```js
data = { name: "Ken", count: null, price: null };
isValid = schema.isValid(data);
// Return true
```

Is there a way to do this easily?

mtth:
You can. Right now the field is nullable but not optional, if you add defaults it'll do what you want:

```javascript
var schema = {
  "type" : "record",
  "name" : "Demo",
  "namespace" : "com.shoprunner.data.demo",
  "fields" : [ {
    "name" : "name",
    "type" : "string"
  }, {
    "name" : "count",
    "type" : [ "null", "int" ],
    "default": null // Here
  }, {
    "name" : "price",
    "type" : [ "null", "float" ],
    "default": null  // And here
  } ]
};
```

loganbfisher:
I am doing the exact same thing as above and am still getting that error.  Any other reason this might not work?

```
{
      "type": "record",
      "name": "CreatedVendorDeployment",
      "namespace": "parkhub.vendor.deployment",
      "doc": "Vendor Deployment Schema",
      "fields": [
        {
          "name": "deploymentId",
          "type": "string"
        },
        {
          "name": "vendorId",
          "type": "string"
        },
        {
          "name": "type",
          "type": "string"
        },
        {
          "name": "collectorId",
          "type": "string"
        },
        {
          "name": "relayId",
          "type": "string"
        },
        {
          "name": "emitterId",
          "type": "string"
        },
        {
          "name": "created",
[...]

PR #265, opened 2019-10-10
Convert undefined to null type when writing

deluxor:
This PR allows the usage of nullable types using the default type undefined, preventing the error when serializing:

`Error: invalid ["null","string"]: undefined`

Basically this converts one to the other.


mtth:
`undefined` and `null` are treated differently by design. To make values optional, you should use [field default values](http://avro.apache.org/docs/current/spec.html#schema_record) (there is no concept of optional outside of records in Avro). https://github.com/mtth/avsc/issues/118 has an example and more information in case you are curious.

PR #282, opened 2020-02-11
Don't add redundant namespaces to types

### LINKED ISSUE #281: Redundant namespace for imported types
Given the following AVDL files:

import.avdl:
```
@namespace("com.org.domain.fixtures")
protocol ImportProto {
  import idl "simple.avdl";

  record Import {
    com.org.domain.fixtures.Simple simple;
  }
}
```

simple.avdl
```
@namespace("com.org.domain.fixtures")
protocol SimpleProto {
  record Simple {
    string foo;
  }
}
```

The resulting AVSC file looks like this:

```
{
    "namespace": "com.org.domain.fixtures",
    "type": "record",
    "name": "Import",
    "fields": [
    {
        "type": {
        "type": "record",
        "name": "Simple",
        "fields": [
            {
            "type": "string",
            "name": "foo"
            }
        ],
        "namespace": "com.org.domain.fixtures"
        },
        "name": "simple"
    }
    ]
}
```

Note how `Simple` has the same namespace as `ImportProto`, yet still has a namespace field. Compare this to the Java equivalent using `avro-tools`:

```
{
    "type": "record",
    "name": "Import",
    "namespace": "com.org.domain.fixtures",
    "fields": [
    {
        "name": "simple",
        "type": {
        "type": "record",
        "name": "Simple",
        "fields": [
            {
            "name": "foo",
            "type": "string"
            }
        ]
        }
    }
    ]
}
```

Nevon:
(Eventually fixes #281)

I believe this works, but I'm not sure what the appropriate behavior is with nested imports. There is a currently failing test that illustrates this well.

If it is correct that we should reset the namespace on any nested imports, the question becomes how to do that with the recursive nature of `assembleProtocol`. I would need to have a reference to the top level protocol.

### DIFF
--- a/lib/specs.js
+++ b/lib/specs.js
@@ -120,12 +120,13 @@ function assembleProtocol(fpath, opts, cb) {
       cb(err);
       return;
     }
+
     // Merge  first the types (where we don't need to check for duplicates
     // since instantiating the service will take care of it), then the messages
     // (where we need to, as duplicates will overwrite each other).
     (importedSchema.types || []).forEach(function (typeSchema) {
-      // Ensure the imported protocol's namespace is inherited correctly (it
-      // might be different from the current one).
+      // Ensure the imported protocol's namespace is inherited correctly if it is different
+      // from the current one.
       if (typeSchema.namespace === undefined) {
         var namespace = importedSchema.namespace;
         if (!namespace) {
@@ -134,7 +135,10 @@ function assembleProtocol(fpath, opts, cb) {
             namespace = match[1];
           }
         }
-        typeSchema.namespace = namespace || '';
+
+        if (namespace !== protocol.namespace && namespace != null) {
+          typeSchema.namespace = namespace
+        }
       }
       importedTypes.push(typeSchema);
     });
--- a/test/test_specs.js
+++ b/test/test_specs.js
@@ -238,8 +238,8 @@ suite('specs', function () {
         assert.strictEqual(err, null);
         assert.deepEqual(schema, {
           protocol: 'First',
-          messages: {two: {request: [], response: 'int'}},
-          types: [{name: 'Foo', type: 'fixed', size: 1, namespace: ''}]
+          messages: { two: { request: [], response: 'int' } },
+          types: [{ name: 'Foo', type: 'fixed', size: 1 }]
         });
         done();
       });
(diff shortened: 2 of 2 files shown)


mtth:
Thanks for the PR, @Nevon.

> If it is correct that we should reset the namespace on any nested imports, the question becomes how to do that with the recursive nature of assembleProtocol. I would need to have a reference to the top level protocol.

Agreed that we should reset the namespace consistently for all imports: it would be surprising for only a subset of imports to have their namespace stripped. It is a bigger change but worth the effort. One way to do it is to distinguish the outermost `opts.importHook` call from the nested ones, stripping the namespaces only after the former.

test/test_specs.js
```diff
@@ -238,13 +238,29 @@ suite('specs', function () {
...
+      var opts = {
+        importHook: createImportHook({
+          '1.avdl': '@namespace("com.example.test")\nprotocol First { record Foo { string foo; } }',
+          '2.avdl': '@namespace("com.example.different")\nprotocol Different { record Baz { string baz; } }',
+          '3.avdl': '@namespace("com.example.test")\nprotocol Second { import idl "1.avdl";\nimport idl "2.avdl";\nrecord Bar { string bar; com.example.test.Foo foo; com.example.different.Baz baz; } }'
+        })
+      }
+      assembleProtocol('3.avdl', opts, function (err, schema) {
+        assert.strictEqual(err, null);
+        assert.strictEqual(schema.types.find(type => type.name === "Baz").namespace, "com.example.different")
```
mtth: The lambdas should be turned to vanilla `function`s. `avsc` is compatible with pre-ES6 versions.

lib/specs.js
```diff
@@ -120,12 +120,13 @@ function assembleProtocol(fpath, opts, cb) {
...
       return;
     }
+
     // Merge  first the types (where we don't need to check for duplicates
     // since instantiating the service will take care of it), then the messages
     // (where we need to, as duplicates will overwrite each other).
     (importedSchema.types || []).forEach(function (typeSchema) {
-      // Ensure the imported protocol's namespace is inherited correctly (it
-      // might be different from the current one).
+      // Ensure the imported protocol's namespace is inherited correctly if it is different
```
mtth: Can we keep the text width consistent?

mtth:
Hi @Nevon. FYI I submitted [a change](https://github.com/mtth/avsc/commit/64179935e31947fa535e757df4c075dbbfac05ad) which addresses this, including nested imports. Thank you for reporting and giving it a shot!

Nevon:
Thanks for fixing it. To be honest, I just got busy with other things and forgot to finish this 🙈

PR #303, opened 2020-05-15
feat: add sample method to generate a predictive sample value of a type

delphinaubin:
Hi and thank you for this nice library !
This pull request adds a `sample()` method to each type which generates a predictive sample value.
It's inspired by the `random()` method but it's more predictable and works better for a schema preview 😄.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -1424,11 +1424,10 @@ UnwrappedUnionType.prototype.random = function () {
   return this.types[index].random();
 };
 
+
 UnwrappedUnionType.prototype.sample = function () {
-  var index = this.types.findIndex(function(type) {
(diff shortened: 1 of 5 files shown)


mtth:
Hi @delphinaubin. Adding methods which are variations of `random` to the type itself unfortunately does not scale well. `random` is one of the early design choices which I would do differently in hindsight: it is better exposed via separate utility functions, which are both more modular and easier to emulate.

As you have noticed there is no random data generation ideal for all use-cases, and I'm afraid including a method per use-case here will lead to an overly large core library. I would encourage you to instead expose `sample` via a separate package, perhaps alongside other sampling methods geared towards other use-cases.

PR #314, opened 2020-08-07
ForSchemaOptions.typeHook should accept undefined

DiogoDoreto:
Hi, I've started migrating my project from JS to TS and got an error due to the fact that the type declaration for `typeHook` option does not accept returning `undefined` values.

According to https://github.com/mtth/avsc/blob/f69d61cdeddaee7ba16b9126e0bcf8eaf0a77026/lib/types.js#L154, returning `undefined` should be fine: code behaves as if the `typeHook` was not provided for that specific type we are going through.


mtth:
Thanks @DiogoDoreto!

PR #316, opened 2020-08-27
permit zero length fixed type as specified by avro

fmsy:
The Avro spec does not restrict the fixed type to have at least 1 byte of length
in fact: the binary encoding ist designed to omit a zero length field completely
using the schema this allows one to annotate types with static information which doesn not have to travel across the wire


mtth:
Thanks for the PR @fmsy!

PR #338, opened 2021-03-23
Performance improvements to ArrayType.

ppershing:
We benchmarked this on internal types with arrays of longer lengths and it brought us ~10% improvement on internal benchmarks.
Note: This creates "holey" arrays in V8 but this seems to have negligible performance impact on the array access speed, presumably JIT or CPU learn to default branch in branch prediction

Sample run:

old:
fromBuffer      | toBuffer      | isValid       | (ops/sec)
---------:      | -------:      | ------:       | ---------
19,109,491      | 5,040,556     | 69,815,976    | ArrayInt.avsc
1,137,653       | 892,341       | 64,718,404    | ArrayString.avsc

new:
fromBuffer      | toBuffer      | isValid       | (ops/sec)
---------:      | -------:      | ------:       | ---------
21,969,180      | 5,099,055     | 70,740,014    | ArrayInt.avsc
1,233,871       | 891,253       | 64,479,021    | ArrayString.avsc

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -1924,7 +1924,7 @@ ArrayType.prototype._check = function (val, flags, hook, path) {
     }
     return false;
   }
-
+  var items = this.itemsType;
   var b = true;
   var i, l, j;
   if (hook) {
@@ -1933,14 +1933,14 @@ ArrayType.prototype._check = function (val, flags, hook, path) {
     path.push('');
     for (i = 0, l = val.length; i < l; i++) {
       path[j] = '' + i;
-      if (!this.itemsType._check(val[i], flags, hook, path)) {
+      if (!items._check(val[i], flags, hook, path)) {
         b = false;
       }
     }
     path.pop();
   } else {
     for (i = 0, l = val.length; i < l; i++) {
-      if (!this.itemsType._check(val[i], flags)) {
+      if (!items._check(val[i], flags)) {
         return false;
       }
     }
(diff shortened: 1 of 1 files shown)


lib/types.js
```diff
@@ -1950,29 +1950,32 @@ ArrayType.prototype._check = function (val, flags, hook, path) {
...
-  var val = [];
+  var val;
   var n;
+  var i = 0
   while ((n = tap.readLong())) {
     if (n < 0) {
       n = -n;
       tap.skipLong(); // Skip size.
     }
+    val = val || new Array(n)
```
mtth, line 1961: Have you tried playing the same trick for subsequent batches? Something like:

```js
if (val) {
  val.length += n;
} else {
  val = new Array(n);
}
```

Intuitively it would give similar benefits to the ones this change provides for the first batch.

lib/types.js
```diff
@@ -1950,29 +1950,32 @@ ArrayType.prototype._check = function (val, flags, hook, path) {
 
 ArrayType.prototype._read = function (tap) {
   var items = this.itemsType;
-  var val = [];
+  var val;
   var n;
+  var i = 0
```
mtth, line 1955: Nit: can you keep the style consistent throughout? For example semi-colons and variable grouping (initialized ones first and uninitialized on a single line after).

```js
var items = this.itemsType;
var i = 0;
var n, val;
```

mtth (commented):
Thanks for the PR @ppershing, the numbers look promising. Do you have benchmark results for different magnitudes of array sizes?

lib/types.js
ppershing: Unfortunately, v8 does not really like `.length` writes. Here is a quick benchmark which demonstrates the fact
https://jsbench.me/g3kmswtm4m/1 and I have seen similar results in other contexts as well

lib/types.js
ppershing: fixed.
Side-note: you should probably setup prettier for the project. I was a bit struggling with it because vscode kept re-formatting the text by default.

lib/types.js
mtth: Interesting, thanks for the benchmark.

mtth:
Thanks for the improvement @ppershing - released in `5.6.1`.

Issue #381, opened 2022-03-22
readProtocol function stuck in infinite loop when doc is empty


jthure:
The readProtocol gets stuck and never finishes if a schema element has an empty doc `/** */`

PR #382, opened 2022-03-22
Allow empty doc in readProtocol

jthure:
Fix issue where readProtocol would get stuck in infinite loop if an schema element has an empty doc

Fixes issue #381

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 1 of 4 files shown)


mtth:
Thanks @jthure. Merged in https://github.com/mtth/avsc/commit/aa2d6c857d93f6425f8ad0bae2ff9ffbb09794ce.

PR #384, opened 2022-03-28
Improved number type inference

nhruo123:
I have noticed that the type inference wasn't super strict, this pull request improves on the number type inference as such:

1. Added a test if the given number is a long
2. Made sure float is only returned if its within a safe value (within the fraction part of the float)

This might be useful with more strict type Avro schema checker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/lib/types.js
+++ b/lib/types.js
@@ -55,6 +55,19 @@ var LOGICAL_TYPE = null;
 // to be able to reference names (i.e. for branches) during instantiation.
 var UNDERLYING_TYPES = [];
 
+
+// range of values represented by 32 bit signed int
+const MAX_INT = 2 ** 31 - 1;
+const MIN_INT = 2 ** 31 * -1;
+
+// range of values represented by 64 bit signed int
+const MAX_LONG = 2 ** 63 - 1;
+const MIN_LONG = 2 ** 63 * -1;
+
+// range of values possible to represent by IEEE 754 single precision floating point fraction part
+const MAX_SAFE_FLOAT = 2 ** 23 - 1;
+const MIN_SAFE_FLOAT = -MAX_SAFE_FLOAT;
+
 /**
  * "Abstract" base Avro type.
  *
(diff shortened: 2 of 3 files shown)


lib/types.js
```diff
@@ -55,6 +55,19 @@ var LOGICAL_TYPE = null;
...
 var UNDERLYING_TYPES = [];
 
+
+// range of values represented by 32 bit signed int
+const MAX_INT = 2 ** 31 - 1;
+const MIN_INT = 2 ** 31 * -1;
+
+// range of values represented by 64 bit signed int
+const MAX_LONG = 2 ** 63 - 1;
+const MIN_LONG = 2 ** 63 * -1;
```
mtth, line 65: This isn't safe unfortunately, see https://github.com/mtth/avsc/wiki/Advanced-usage#custom-long-types. Here you can use `isSafeLong` instead.

lib/types.js
```diff
@@ -252,11 +265,19 @@ Type.forValue = function (val, opts) {
     case 'boolean':
       return Type.forSchema('boolean', opts);
     case 'number':
-      if ((val | 0) === val) {
-        return Type.forSchema('int', opts);
-      } else if (Math.abs(val) < 9007199254740991) {
-        return Type.forSchema('float', opts);
+      if (Number.isInteger(val)) {
+        if (MIN_INT <= val && val <= MAX_INT) {
```
mtth, line 269: Let's replace with `(val | 0) === val`.

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
mtth, line 1: Please omit, would rather avoid having a file just for this folder since it should be in a global ignore file.

lib/types.js
```diff
@@ -55,6 +55,19 @@ var LOGICAL_TYPE = null;
...
+const MAX_INT = 2 ** 31 - 1;
+const MIN_INT = 2 ** 31 * -1;
+
+// range of values represented by 64 bit signed int
+const MAX_LONG = 2 ** 63 - 1;
+const MIN_LONG = 2 ** 63 * -1;
+
+// range of values possible to represent by IEEE 754 single precision floating point fraction part
+const MAX_SAFE_FLOAT = 2 ** 23 - 1;
+const MIN_SAFE_FLOAT = -MAX_SAFE_FLOAT;
```
mtth, line 69: This is misleading - floats can also underflow faster than doubles. This is not a safe range. I'm not sure there's a good way to efficiently and correctly infer float vs double though. I recommend we either keep the existing code as-is or (better but optional) simplify to always inferring double on non-integer and unsafe-long input.

lib/types.js
```diff
@@ -252,11 +265,19 @@ Type.forValue = function (val, opts) {
...
+        } else if (MIN_LONG <= val && val <= MAX_LONG) {
+          return Type.forSchema('long', opts);
+        }
+      } else {
+        if (MIN_SAFE_FLOAT <= val && val <= MAX_SAFE_FLOAT) {
+          return Type.forSchema('float', opts);
+        }
       }
+      
+      // if all else fails, return double its the broadest type.
```
mtth, line 280: Let's omit this comment, it doesn't bring much value and contains a typo.

nhruo123:
Thanks for reviewing, I'll close this issue as it's not as good of an improvement as I initially thought 😅

PR #402, opened 2022-08-13
Minor type definition update

arminhaghi:
Looks like type definition update was missed when https://github.com/mtth/avsc/pull/261 change was submitted.

 Added default value for enum type definition


mtth:
Thanks @arminhaghi.

PR #404, opened 2022-09-19
Make the "main" property point to the "index.js" file

FredericEspiau:
The "main" property of the "package.json" file should point to a "js" file, not to a folder, you can see some examples here: https://github.com/npm/cli/blob/latest/workspaces/arborist/test/arborist/reify.js

I'm surprised nobody complained before, it causes some issues in my projects

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "lib",
     "types"
   ],
-  "main": "./lib",
+  "main": "./lib/index.js",
   "types": "./types",
   "browser": {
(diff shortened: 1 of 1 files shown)


mtth:
Hi @FredericEspiau. Thanks for the PR. Before proceeding, I'd like to understand what is special about your setup. Can you share a way to reproduce the issue? As you point out, `avsc` seems to be importing fine for everyone else. AFAIK it suffices for `main` to point to something that can be used with `require` or `import` (both of which work with folders).

PR #405, opened 2022-09-26
Add buffer import to simplify browser builds

### LINKED ISSUE #403: global Buffer usage complicates browser usage
Accessing Buffer as a global works great in Node.js, but in the browser requires each build tool to inject that global in its own special way.

Would you be open to a small PR that uses import/require to access Buffer to simplify browser build issues?

JAForbes:
Fixes #403

### DIFF
--- a/etc/browser/lib/crypto.js
+++ b/etc/browser/lib/crypto.js
@@ -12,6 +12,9 @@
  *
  */
 
+var buffer = require('buffer');
+var Buffer = buffer.Buffer;
+
 function createHash(algorithm) {
   if (algorithm !== 'md5') {
     throw new Error('only md5 is supported in the browser');
--- a/lib/containers.js
+++ b/lib/containers.js
@@ -15,11 +15,12 @@
 
 var types = require('./types'),
     utils = require('./utils'),
+    buffer = require('buffer'),
     stream = require('stream'),
     util = require('util'),
     zlib = require('zlib');
 
-
+var Buffer = buffer.Buffer;
 var OPTS = {namespace: 'org.apache.avro.file'};
 
 var LONG_TYPE = types.Type.forSchema('long', OPTS);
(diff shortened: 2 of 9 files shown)


afzaal-ameer:
@JAForbes I haven't tested this, but I see references to both `buffer` and `Buffer` in some files. e.g. https://github.com/JAForbes/avsc/blob/52d2b93be59af6f5228599399c0881aad25e70e0/lib/index.js#L69

etc/benchmarks/avro-serialization-implementations/scripts/decode/node-avro-io.js
```diff
@@ -5,8 +5,8 @@
 'use strict';
 
 var io = require('node-avro-io'),
-    avsc = require('../../../../lib');
-
+    avsc = require('../../../../lib'),
+    {Buffer} = require('buffer');
```
mtth, line 9: This syntax isn't compatible with older Node versions, could you swap to one that is? For example:

```js
    // ...
    buffer = require('buffer');

var Buffer = buffer.Buffer;
```

Also, feel free to skip the files in `etc/` as they are not included in the distributed package.

mtth (changes requested):

etc/benchmarks/avro-serialization-implementations/scripts/decode/node-avro-io.js
JAForbes: Yep no worries.  Yeah I think that syntax is Node 16+.

etc/browser/lib/crypto.js
```diff
@@ -12,6 +12,9 @@
  *
  */
 
+var buffer = require('buffer');
+var Buffer = buffer.Buffer;
```
JAForbes, line 16: @mtth I left this /etc file in, I wasn't sure if it was a polyfill for the browser and may need the explicit import.

mtth: Good call, it is. I'd forgotten about that one.

mtth (approved):
Thanks for the PR, @JAForbes.

mtth:
Published as `5.7.6`.

PR #428, opened 2023-01-10
Bump version and omit services

mtth:
Also delete all previously deprecated functionality.

Wiki documentation (API, ...) will be updated separately once this is in.


valadaptive:
I'll try to finish the schema path resolution abstraction either today or tomorrow (removing the `fs` dependency from browser distributions) and submit a PR. That's the only breaking change I can think of.

PR #445, opened 2024-01-14
Refactor import hook API to subsume all platform-dependent behavior

valadaptive:
First off, apologies for leaving you hanging for a year! Some stuff came up and this fell off my radar for a while, and this seemed like it was going to be more fiddly and difficult than it ended up being.

This import hook API isolates all platform-specific path resolution stuff within the import hook itself. The hook now takes the path of the imported file *and* the path of the file that imported it, and calls the callback with the file contents, and the *resolved* path of the imported file. That second part is what stumped me on my last attempt.

I also started work last year on a "de-Buffer-ification" branch, but ran into some performance cliffs. I finally figured those out (except for TextEncoder/TextDecoder, which is still slower in Node sadly), and the code [is now on GitHub](https://github.com/valadaptive/avsc/tree/debufferify). It's actually *faster* on most benchmarks than the `Buffer` version. It's still a proof-of-concept and will need to be cleaned up. If you're interested in merging it *before* the major version bump, maybe remove `services` first, so I don't have to port that over to use `Uint8Array`.

### DIFF
--- a/etc/browser/avsc-services.js
+++ b/etc/browser/avsc-services.js
@@ -7,7 +7,6 @@
  */
 
 let avroTypes = require('./avsc-types'),
-    files = require('../../lib/files'),
     services = require('../../lib/services'),
     specs = require('../../lib/specs'),
     utils = require('../../lib/utils');
@@ -15,7 +14,7 @@ let avroTypes = require('./avsc-types'),
 
 /** Slightly enhanced parsing, supporting IDL declarations. */
 function parse(any, opts) {
-  let schemaOrProtocol = files.readSchemaFromPathOrString(any);
+  let schemaOrProtocol = specs.read(any);
   return schemaOrProtocol.protocol ?
     services.Service.forProtocol(schemaOrProtocol, opts) :
     avroTypes.Type.forSchema(schemaOrProtocol, opts);
--- a/etc/browser/avsc-services.js
+++ b/etc/browser/avsc-services.js
@@ -7,6 +7,7 @@
  */
 
 let avroTypes = require('./avsc-types'),
+    files = require('../../lib/files'),
     services = require('../../lib/services'),
     specs = require('../../lib/specs'),
     utils = require('../../lib/utils');
@@ -14,7 +15,7 @@ let avroTypes = require('./avsc-types'),
 
 /** Slightly enhanced parsing, supporting IDL declarations. */
 function parse(any, opts) {
-  let schemaOrProtocol = specs.read(any);
+  let schemaOrProtocol = files.readSchemaFromPathOrString(any);
   return schemaOrProtocol.protocol ?
     services.Service.forProtocol(schemaOrProtocol, opts) :
     avroTypes.Type.forSchema(schemaOrProtocol, opts);
--- a/etc/browser/lib/files.js
+++ b/etc/browser/lib/files.js
@@ -12,10 +12,12 @@ function createSyncImportHook() {
   return function () { throw createError(); };
 }
 
+function tryReadFileSync() { return null; }
+
 module.exports = {
   createImportHook,
   createSyncImportHook,
   existsSync: function () { return false; },
   readFileSync: function () { throw createError(); },
-  readSchemaFromPathOrString: function () { throw createError(); }
+  tryReadFileSync,
 };
--- a/etc/browser/lib/files.js
+++ b/etc/browser/lib/files.js
@@ -17,7 +17,5 @@ function tryReadFileSync() { return null; }
 module.exports = {
   createImportHook,
   createSyncImportHook,
-  existsSync: function () { return false; },
-  readFileSync: function () { throw createError(); },
   tryReadFileSync,
 };
--- a/etc/browser/lib/files.js
+++ b/etc/browser/lib/files.js
@@ -5,7 +5,7 @@
 function createError() { return new Error('unsupported in the browser'); }
 
 function createImportHook() {
-  return function (fpath, kind, cb) { cb(createError()); };
+  return function (_, cb) { cb(createError()); };
 }
 
 function createSyncImportHook() {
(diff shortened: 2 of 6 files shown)


mtth:
No worries and thanks for the PR @valadaptive! This change looks good overall, will send a review shortly.

Great to hear about the `Uint8Array`-related changes. I would probably release those with the upcoming major version. Could you share performance numbers?

lib/files.js
```diff
@@ -13,15 +13,18 @@ let fs = require('fs'),
 /** Default (asynchronous) file loading function for assembling IDLs. */
 function createImportHook() {
   let imports = {};
-  return function (fpath, kind, cb) {
-    fpath = path.resolve(fpath);
+  return function ({path: fpath, parentPath}, cb) {
```
mtth, line 16: Can we rename to `importerPath` to be more explicit?

Also curious if you considered adding the importer's _parent directory_'s path directly here.

lib/specs.js
```diff
@@ -193,7 +196,7 @@ function assembleProtocol(fpath, opts, cb) {
  *
  * The parsing logic is as follows:
  *
- * + If `str` contains `path.sep` (on windows `\`, otherwise `/`) and is a path
```
mtth, line 196: Can you elaborate on this change?

types/index.d.ts
```diff
@@ -127,8 +127,18 @@ interface IsValidOptions {
   noUndeclaredFields: boolean;
   errorHook: (path: string[], val: any, type: Type) => void
 }
+
+interface ImportHookParams {
```
mtth, line 131: "Params" doesn't feel right for a hook since the data is provided to (not by) the caller. How about `ImportHookData` or `ImportHookPayload`?

lib/specs.js
mtth: > I could try to move this function into some Node-specific module or exclude it from the browser build, so that path.sep can be used--does that sound like a better route?

That sounds great. Could you do it in `lib/files.js`?

mtth:
Thanks @valadaptive. The numbers above look great. Did you mean to push changes to this branch?

valadaptive:
Yup, my bad! Changes have been properly pushed now.

lib/files.js
```diff
@@ -35,22 +39,90 @@ function createImportHook() {
...
+  let schema;
+  if (
+    typeof str == 'string' &&
+    str.indexOf('/') !== -1 &&
+    fs.existsSync(str)
+  ) {
+    // Try interpreting `str` as path to a file contain a JSON schema or an IDL
+    // protocol. Note that we add the second check to skip primitive references
+    // (e.g. `"int"`, the most common use-case for `avro.parse`).
+    let contents = fs.readFileSync(str, {encoding: 'utf8'});
```
mtth, line 88: This is the only part that requires `fs` - could we trim this function only include this logic? We could call it something like `tryReadFileSync`. The change would avoid the `./spec` import here and logic duplication between the two files.

lib/files.js
```diff
@@ -35,22 +39,90 @@ function createImportHook() {
...
+ *   if no imports are present (and an error is thrown if there are any
+ *   imports).
+ * + If `str` is a valid IDL type specification, it is parsed and returned.
+ * + If neither of the above cases apply, `str` is returned.
+ */
+function readSchemaFromPathOrString(str) {
+  let schema;
+  if (
+    typeof str == 'string' &&
+    str.indexOf('/') !== -1 &&
```
mtth, line 82: `path.sep`?

mtth (commented):
Apologies for the slow review.

lib/files.js
valadaptive: Done in the latest commit.

etc/browser/avsc-services.js
```diff
@@ -7,14 +7,15 @@
...
+    files = require('../../lib/files'),
     services = require('../../lib/services'),
     specs = require('../../lib/specs'),
     utils = require('../../lib/utils');
 
 
 /** Slightly enhanced parsing, supporting IDL declarations. */
 function parse(any, opts) {
-  let schemaOrProtocol = specs.read(any);
+  let schemaOrProtocol = files.readSchemaFromPathOrString(any);
```
mtth, line 18: Revert now that `files` has been updated?

etc/browser/lib/files.js
```diff
@@ -5,17 +5,20 @@
...
+function tryReadFileSync() { return null; }
 
 module.exports = {
   createImportHook,
   createSyncImportHook,
   existsSync: function () { return false; },
-  readFileSync: function () { throw createError(); }
+  readFileSync: function () { throw createError(); },
+  tryReadFileSync,
+  readSchemaFromPathOrString: function () { throw createError(); },
```
mtth, line 23: Omit this line?

lib/files.js
```diff
@@ -35,22 +38,42 @@ function createImportHook() {
...
+ * (`path.sep`). If not, this will return `null`.
+ */
+function tryReadFileSync(str) {
+  if (
+    typeof str == 'string' &&
+    str.indexOf(path.sep) !== -1 &&
+    fs.existsSync(str)
+  ) {
+    // Try interpreting `str` as path to a file.
+    return fs.readFileSync(str, {encoding: 'utf8'});
```
mtth, line 67: Optional: the `existsSync` call could be omitted by catching `readFileSync` errors and checking the error's code.

mtth (commented):
Thanks for updating. Looks good - just two minor changes and this should be ready to merge.

etc/browser/avsc-services.js
valadaptive: Good catch; reverted

lib/files.js
valadaptive: I've done so

valadaptive:
I've made those last few changes and also stopped exporting existsSync and readFileSync now that that functionality's been moved into files.js.

mtth (commented):
Thanks @valadaptive!

Issue #442, opened 2023-10-09
Support ?-syntax for optional fields in avdl


adamlindqvist:
I'm trying to parse a avdl file with an optional field but it seems that `avsc` does not support using the `?`-syntax to describe a optional/nullable field. 

Example

.avdl file 
```avdl
protocol Protocol {
	record Record {
		string? text;
	}
}
```

Code:
```ts
const path = path.resolve('./path-to-avdl-file.avdl');
const schema = fs.readFileSync(path, { encoding: 'utf8' });
const protocol = avsc.readProtocol(schema);
```

Resulting in error:
```
Error: invalid token {"pos":47,"id":"operator","val":"?"}: expected ID name
```

Is this a known issue or is it possible to work around this?

_Context: https://avro.apache.org/docs/1.11.1/idl-language/#unions_

mtth:
Hi @adamlindqvist. Thanks for reporting - the optional shorthand isn't currently supported (PR very welcome!).

You'll need to stick to the long form until it is:

```avdl
protocol Protocol {
	record Record {
		union { null, string } text;
	}
}
```

yehonatanz:
@mtth Since the corresponding PR is now merged, any chance to release a new version to NPM to make this feature accessible?

PR #446, opened 2024-01-16
Support optional fields using question mark syntax

yehonatanz:
Addresses #442

### DIFF
--- a/lib/specs.js
+++ b/lib/specs.js
@@ -427,10 +427,14 @@ class Reader {
     if (javadoc !== undefined && schema.doc === undefined) {
       schema.doc = javadoc;
     }
+    const isOptional = tk.next({id: 'operator', val: '?', silent: true});
     this._readAnnotations(schema);
     schema.name = tk.next({id: 'name'}).val;
     if (tk.next({val: '=', silent: true})) {
       schema['default'] = tk.next({id: 'json'}).val;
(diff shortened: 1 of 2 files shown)


lib/specs.js
```diff
@@ -463,7 +463,9 @@ class Reader {
           delete schema.type; // Always overwrite the type.
           utils.copyOwnProperties(ref, schema);
         }
-        return Object.keys(schema).length > 1 ? schema : schema.type;
+        const isOptional = this._tk.next({id: 'operator', val: '?', silent: true});
+        const result = Object.keys(schema).length > 1 ? schema : schema.type;
+        return isOptional ? ['null', result]: result;
```
mtth, line 468: The position of the `null` may not be right here: it should be the second element if the field has a non-null default value. See https://avro.apache.org/docs/1.11.1/idl-language/#unions for context. You could address this by moving this logic to `_readField` (arguably better also since optionality is a field-level concept rather than type) and using the default's value (if any) to determine the right position.

mtth (changes requested):
Thanks for the PR @yehonatanz!

lib/specs.js
yehonatanz: Oh, I missed that, thank you!
Fixed my code accordingly :)

PR #467, opened 2024-09-10
fix(types): `forSchema` works with arrays (union)

joscha:
See https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/types.js#L190

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -189,7 +189,7 @@ export class Type {
   readonly name: string | undefined;
   readonly branchName: string | undefined;
   readonly typeName: string;
-  static forSchema(schema: Schema, opts?: Partial<ForSchemaOptions>): Type;
+  static forSchema(schema: Schema | Schema[], opts?: Partial<ForSchemaOptions>): Type;
   static forTypes(types: Type[], opts?: Partial<TypeOptions>): Type;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -189,7 +189,7 @@ export class Type {
   readonly name: string | undefined;
   readonly branchName: string | undefined;
   readonly typeName: string;
-  static forSchema(schema: Schema, opts?: Partial<ForSchemaOptions>): Type;
+  static forSchema(schema: Schema | Schema[], opts?: Partial<ForSchemaOptions>): Type;
```
mtth, line 192: Unfortunately this will cause `forSchema` to accept invalid inputs. `Schema` already contains array types (`DefinedType[]` and `Type[]`), and nested arrays should not be allowed.

joscha: My code had:
```
Avro.Type.forSchema([Avro.Type.forSchema('int'), Avro.Type.forSchema('string')])
```
which works with the type change but fails with the current definitions due to `Type` not assignable to `DefinedType`.

I think this should have been:
```
Avro.Type.forTypes([Avro.Type.forSchema('int'), Avro.Type.forSchema('string')])
```
instead, possibly?

Issue #239, opened 2019-06-14
Help with unions of complex types?


murkey:
Hello! I'm trying to implement a union type and running into an error. It seems like others are having some success (https://github.com/mtth/avsc/issues/174) so I'm probably just missing something, but any help would be much appreciated.

Code to reproduce the issue:
```js
const avro = require('avsc');

const type = avro.Type.forSchema({
  type: 'record',
  fields: [{
    name: 'blah',
    type: [
      {type: 'record', name: 'union1', fields: [{name: 'foo', type: 'string'}]},
      {type: 'record', name: 'union2', fields: [{name: 'bar', type: 'int'}]}
    ]
  }]
});

const buf = type.toBuffer({blah: {foo: 'baz'}});
const val = type.fromBuffer(buf);
console.log(val)
```

Error:
```
/Users/murkey/code/avro-test/node_modules/avsc/lib/types.js:3017
  throw new Error(f('invalid %j: %j', type.schema(), val));
  ^

Error: invalid [{"name":"union1","type":"record","fields":[{"name":"foo","type":"string"}]},{"name":"union2","type":"record","fields":[{"name":"bar","type":"int"}]}]: {"foo":"baz"}
    at throwInvalidError (/Users/murkey/code/avro-test/node_modules/avsc/lib/types.js:3017:9)
    at WrappedUnionType._write (/Users/murkey/code/avro-test/node_modules/avsc/lib/types.js:1496:7)
    at RecordType.writeRecord$ [as _write] (eval at RecordType._createWriter (/Users/murkey/code/avro-test/node_modules/avsc/lib/types.js:2312:10), <anonymous>:4:6)
    at RecordType.Type.toBuffer (/Users/murkey/code/avro-test/node_modules/avsc/lib/types.js:646:8)
    at Object.<anonymous> (/Users/murkey/code/avro-test/index.js:14:18)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
```

I've also tried wrapping the union schemas with `avro.Type.forSchema`:
```js
    type: [
      avro.Type.forSchema({type: 'record', name: 'union1', fields: [{name: 'foo', type: 'string'}]}),
      avro.Type.forSchema({type: 'record', name: 'union2', fields: [{name: 'bar', type: 'int'}]})
    ]
```
But this produced the same error.

[...]

PR #469, opened 2024-09-11
feat: allow named types in unions

joscha:
The Avro specification defines:

> Unions may not contain more than one schema with the same type, except for the named types record, fixed and enum

see https://avro.apache.org/docs/1.11.1/specification/#unions

meaning that if a `record` type has a name, it is valid to mix multiple of them in a union, unwrapped.

The same change should possibly be made for enums as well.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -1252,28 +1252,44 @@ function generateProjectionIndexer(projectionFn) {
   };
 }
 
-function generateDefaultIndexer() {
-  this._dynamicBranches = null;
-  this._bucketIndices = {};
-  this.types.forEach(function (type, index) {
-    if (Type.isType(type, 'abstract', 'logical')) {
-      if (!this._dynamicBranches) {
-        this._dynamicBranches = [];
+function generateDefaultIndexer(types) {
+  const dynamicBranches = [];
+  const bucketIndices = {};
+
+  const getBranchIndex = (any, index) => {
+    let logicalBranches = dynamicBranches;
+    for (let i = 0, l = logicalBranches.length; i < l; i++) {
+      let branch = logicalBranches[i];
+      if (branch.type._check(any)) {
+        if (index === undefined) {
+          index = branch.index;
+        } else {
+          // More than one branch matches the value so we aren't guaranteed to
+          // infer the correct type. We throw rather than corrupt data. This can
+          // be fixed by "tightening" the logical types.
+          throw new Error('ambiguous conversion');
+        }
       }
-      this._dynamicBranches.push({index, type});
+    }
+    return index;
+  }
+
+  types.forEach(function (type, index) {
+    if (Type.isType(type, 'abstract', 'logical')) {
+      dynamicBranches.push({index, type});
     } else {
       let bucket = getTypeBucket(type);
-      if (this._bucketIndices[bucket] !== undefined) {
+      if (bucketIndices[bucket] !== undefined) {
         throw new Error(`ambiguous unwrapped union: ${j(this)}`);
       }
-      this._bucketIndices[bucket] = index;
+      bucketIndices[bucket] = index;
     }
   }, this);
   return (val) => {
-    let index = this._bucketIndices[getValueBucket(val)];
-    if (this._dynamicBranches) {
+    let index = bucketIndices[getValueBucket(val)];
+    if (dynamicBranches.length) {
       // Slower path, we must run the value through all branches.
-      index = this._getBranchIndex(val, index);
+      index = getBranchIndex(val, index);
     }
     return index;
   };
@@ -1310,28 +1326,10 @@ class UnwrappedUnionType extends UnionType {
       }
     }
     this._getIndex = _projectionFn
-      ? generateProjectionIndexer(projectionFn)
+      ? generateProjectionIndexer(_projectionFn)
       : generateDefaultIndexer.bind(this)(this.types);
 
     Object.freeze(this);
-  }
-
-  _getBranchIndex (any, index) {
-    let logicalBranches = this._dynamicBranches;
-    for (let i = 0, l = logicalBranches.length; i < l; i++) {
-      let branch = logicalBranches[i];
-      if (branch.type._check(any)) {
-        if (index === undefined) {
-          index = branch.index;
-        } else {
-          // More than one branch matches the value so we aren't guaranteed to
-          // infer the correct type. We throw rather than corrupt data. This can
-          // be fixed by "tightening" the logical types.
-          throw new Error('ambiguous conversion');
-        }
-      }
-    }
-    return index;
   }
 
   _check (val, flags, hook, path) {
@@ -1393,16 +1391,18 @@ class UnwrappedUnionType extends UnionType {
           // Using the `coerceBuffers` option can cause corruption and erroneous
           // failures with unwrapped unions (in rare cases when the union also
           // contains a record which matches a buffer's JSON representation).
-          if (isJsonBuffer(val) && this._bucketIndices.buffer !== undefined) {
-            index = this._bucketIndices.buffer;
-          } else {
-            index = this._getIndex(val);
+          if (isJsonBuffer(val)) {
+            let bufIndex = this.types.findIndex(t => getTypeBucket(t) === 'buffer');
+            if (bufIndex !== -1) {
+              index = bufIndex;
+            }
           }
+          index ??= this._getIndex(val);
           break;
         case 2:
           // Decoding from JSON, we must unwrap the value.
           if (val === null) {
-            index = this._bucketIndices['null'];
+            index = this._getIndex(null);
           } else if (typeof val === 'object') {
             let keys = Object.keys(val);
             if (keys.length === 1) {
--- a/lib/types.js
+++ b/lib/types.js
@@ -67,6 +67,7 @@ class Type {
     this.name = undefined;
     this.aliases = undefined;
     this.doc = (schema && schema.doc) ? '' + schema.doc : undefined;
+    this.opts = Object.assign({}, opts);
 
     if (schema) {
       // This is a complex (i.e. non-primitive) type.
@@ -351,7 +352,7 @@ class Type {
     // Group types by category, similar to the logic for unwrapped unions.
     let bucketized = {};
     expanded.forEach((type) => {
-      let bucket = getTypeBucket(type);
+      let bucket = getTypeBucket(type, true);
       let bucketTypes = bucketized[bucket];
       if (!bucketTypes) {
         bucketized[bucket] = bucketTypes = [];
@@ -1258,7 +1259,7 @@ class UnwrappedUnionType extends UnionType {
   }
 
   _getIndex (val) {
-    let index = this._bucketIndices[getValueBucket(val)];
+    let index = this._bucketIndices[getValueBucket(val, this.opts.namespace)];
     if (this._dynamicBranches) {
       // Slower path, we must run the value through all branches.
       index = this._getBranchIndex(val, index);
@@ -3027,8 +3028,9 @@ function maybeQualify(name, ns) {
  * Get a type's bucket when included inside an unwrapped union.
  *
  * @param type {Type} Any type.
+ * @param isExpanded {Boolean} Whether the union is expanded - if yes, we don't use the name to determine the bucket, as the parent type is already in the right bucket.
  */
-function getTypeBucket(type) {
+function getTypeBucket(type, isExpanded) {
   let typeName = type.typeName;
   switch (typeName) {
     case 'double':
@@ -3045,7 +3047,11 @@ function getTypeBucket(type) {
     case 'error':
       return 'object';
     case 'record':
-      return type.name ? maybeQualify(type.name, type.namespace) : 'object';
+      if (isExpanded || !type.name) {
+        return 'object';
+      } else {
+        return maybeQualify(type.name, type.namespace)
+      }
     default:
       return typeName;
   }
@@ -3055,8 +3061,9 @@ function getTypeBucket(type) {
  * Infer a value's bucket (see unwrapped unions for more details).
  *
  * @param val {...} Any value.
+ * @param namespace {String} Optional namespace.
  */
-function getValueBucket(val) {
+function getValueBucket(val, namespace) {
   if (val === null) {
     return 'null';
   }
@@ -3067,6 +3074,12 @@ function getValueBucket(val) {
       return 'array';
     } else if (Buffer.isBuffer(val)) {
       return 'buffer';
+    } else if(typeof val.constructor !== 'undefined' && val.constructor.name) {
+      const isBuiltin = Object.values(module.exports.builtins).some(b => val instanceof b);
+      if (isBuiltin || val.constructor === Object) {
+        return bucket;
+      }
+      return maybeQualify(val.constructor.name, namespace);
     }
   }
   return bucket;
--- a/lib/types.js
+++ b/lib/types.js
@@ -3044,7 +3044,7 @@ function getTypeBucket(type) {
     case 'map':
     case 'error':
     case 'record':
-      return 'object';
+      return type.name ? maybeQualify(type.name, type.namespace) : 'object';
     default:
       return typeName;
   }
--- a/lib/types.js
+++ b/lib/types.js
@@ -3076,9 +3076,8 @@ function getTypeBucket(type) {
       return 'string';
     case 'map':
     case 'error':
-      return 'object';
     case 'record':
-      return type.name ? maybeQualify(type.name, type.namespace) : 'object';
+      return 'object';
     default:
       return typeName;
   }
--- a/lib/types.js
+++ b/lib/types.js
@@ -2987,10 +2987,11 @@ function readArraySize(tap) {
  *
  * + We are not using the `Number` constants for compatibility with older
  *   browsers.
- * + We must remove one from each bound because of rounding errors.
+ * + We divide the bounds by two to avoid rounding errors during zigzag encoding
+ *   (see https://github.com/mtth/avsc/issues/455).
  */
 function isSafeLong(n) {
-  return n >= -9007199254740990 && n <= 9007199254740990;
+  return n >= -4503599627370496 && n <= 4503599627370496;
 }
 
 /**
--- a/lib/types.js
+++ b/lib/types.js
@@ -3043,6 +3043,7 @@ function getTypeBucket(type) {
       return 'string';
     case 'map':
     case 'error':
+      return 'object';
     case 'record':
       return type.name ? maybeQualify(type.name, type.namespace) : 'object';
     default:
--- a/lib/types.js
+++ b/lib/types.js
@@ -202,14 +202,10 @@ class Type {
       if (!UnionType) {
         if (typeof opts.wrapUnions === 'function') {
           // we have a projection function
-          try {
-            projectionFn = opts.wrapUnions(types);
-            UnionType = typeof projectionFn !== 'undefined'
-              ? UnwrappedUnionType
-              : WrappedUnionType;
-          } catch(e) {
-            throw new Error(`Error generating projection function: ${e}`);
-          }
+          projectionFn = opts.wrapUnions(types);
+          UnionType = typeof projectionFn !== 'undefined'
+            ? UnwrappedUnionType
+            : WrappedUnionType;
         } else {
           UnionType = isAmbiguous(types) ? WrappedUnionType : UnwrappedUnionType;
         }
@@ -1319,11 +1315,7 @@ class UnwrappedUnionType extends UnionType {
     super(schema, opts);
 
     if (!_projectionFn && opts && typeof opts.wrapUnions === 'function') {
-      try {
-        _projectionFn = opts.wrapUnions(this.types);
-      } catch(e) {
-        throw new Error(`Error generating projection function: ${e}`);
-      }
+      _projectionFn = opts.wrapUnions(this.types);
     }
     this._getIndex = _projectionFn
       ? generateProjectionIndexer(_projectionFn)
--- a/lib/types.js
+++ b/lib/types.js
@@ -1280,7 +1280,7 @@ function generateDefaultIndexer(types) {
       }
       bucketIndices[bucket] = index;
     }
-  }, this);
+  });
   return (val) => {
     let index = bucketIndices[getValueBucket(val)];
     if (dynamicBranches.length) {
@@ -1319,7 +1319,7 @@ class UnwrappedUnionType extends UnionType {
     }
     this._getIndex = _projectionFn
       ? generateProjectionIndexer(_projectionFn)
-      : generateDefaultIndexer.bind(this)(this.types);
+      : generateDefaultIndexer(this.types);
 
     Object.freeze(this);
   }
--- a/lib/types.js
+++ b/lib/types.js
@@ -1266,13 +1266,31 @@ class UnwrappedUnionType extends UnionType {
 
     this._dynamicBranches = null;
     this._bucketIndices = {};
+
+    this.projectionFunction = (val) => this._bucketIndices[getValueBucket(val)];
+    let hasWrapUnionsFn = opts && typeof opts.wrapUnions === 'function';
+    if (hasWrapUnionsFn) {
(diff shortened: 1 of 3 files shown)


lib/types.js
```diff
@@ -3044,7 +3044,7 @@ function getTypeBucket(type) {
     case 'map':
     case 'error':
```
joscha, line 3045:
```suggestion
    case 'error':
      return 'object';
```

joscha:
Closing this for now, I think it needs an additional fix in the bucket selection for `getValueBucket`, see https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/types.js#L1261

joscha:
For my current project's, very specific use-case (avro schemas generated from an openapi definition and then the data queried via a generated client as well, so the class names match) this version works:

```js
function getValueBucket(val) {
  if (val === null) {
    return 'null';
  }
  var bucket = typeof val;
  if (bucket === 'object') {
    // Could be bytes, fixed, array, map, or record.
    if (Array.isArray(val)) {
      return 'array';
    } else if (Buffer.isBuffer(val)) {
      return 'buffer';
    }
    if (val.constructor.name) {
      return 'model.' + val.constructor.name;
    };
  }
  return bucket;
}
```

however
```
    if (val.constructor.name) {
      return 'model.' + val.constructor.name;
    };
```
doesn't transfer to raw JSON. The current namespace would need to be passed into the `UnwrappedUnion` class as well, in order to qualify the type.

Generally, there is a discriminator (an attribute named `type`) on the containing object and the object to serialize, which can be used to discriminate, but that would need a lot more introspection on the types. I wonder if a custom resolver in `getValueBucket` would be acceptable, so the serialization implementation can decide? Thoughts?

lib/types.js
```diff
@@ -3066,6 +3074,12 @@ function getValueBucket(val) {
       return 'array';
     } else if (Buffer.isBuffer(val)) {
       return 'buffer';
+    } else if(typeof val.constructor !== 'undefined' && val.constructor.name) {
```
joscha, line 3077:
```suggestion
    } else if (typeof val.constructor !== 'undefined' && val.constructor.name) {
```

joscha:
Reopened this with the proper implementation respecting the namespace and inferred arrays holding objects. Would be great to get some feedback on this and see if this is the right direction. If yes, will add tests of course. All current tests pass with this change.

mtth:
Unions of named types are already supported using wrapped unions. Can you share some context on your use-case to understand why they aren't a good fit?

Note also that you can already force an unwrapped representation for arbitrary unions using logical types (see [here](https://github.com/mtth/avsc/issues/239#issuecomment-502312583), and the linked comment).

joscha:
> Unions of named types are already supported using wrapped unions. Can you share some context on your use-case to understand why they aren't a good fit?

Of course. I am using this [API from Affinity CRM](https://developer.affinity.co/) (OpenAPI spec at the top or a current extract [here](https://github.com/planet-a-ventures/affinity-node/blob/main/openapi/2024-09-05.json)) to generate a Typescript client with [openapi-generator/typescript](https://openapi-generator.tech/docs/generators/typescript). You can see the implementation of it [here](https://github.com/planet-a-ventures/affinity-node/tree/main/src/v2/generated).

Our current Affinity CRM instance has multiple 10 thousand items in it of the `Company` type.
A `Company` type can have an arbitrary number of fields with each field having one of many different value types (see [here](https://github.com/planet-a-ventures/affinity-node/blob/364a0b2ac2f86d04b128f8ebb0ac8560f36ab8dd/src/v2/generated/models/FieldValue.ts#L36)). The overall data is about 1.4 GB of raw JSON.

I am using [openapi-generator/avro-schema](https://openapi-generator.tech/docs/generators/avro-schema/) to generate an Avro schema for the same data.

I then read the data in batches of 100 from the API and stream them directly through `avsc` on disk using a file encoder.
From there I import them into [Snowflake](https://www.snowflake.com).

I want the data in Snowflake to be:
- as simple as possible
- as close as possible to the OpenAPI response from the API

I also want to:
- ideally not having to post-process the data before sending it to avsc; wrapped unions would require that, as I need to change the JSON data passed to avsc to something like
  ```js
  field.value = {
    'model.TextValue': field.value
  }
  ```
  in order to adhere to the rule that the value needs to sit in an object with a single key of its type name
- not change the automatically generated Avro schema to resemble wrapped unions as avsc understands them for two reasons:
  1. I don't want to bind the data structure / Avro schema to the current implementation of avsc
  2. I don't want to have to apply a patch (bad) or manual changes (even worse) to the generated schema each time the OpenAPI spec from the API gets an update.

Does that make sense?
 
> Note also that you can already force an unwrapped representation for arbitrary unions using logical types (see [here](https://github.com/mtth/avsc/issues/239#issuecomment-502312583), and the linked comment).

Yes, I am currently using `wrapUnions: 'never'`, however the unions I have do almost certainly not need wrapping. I believe the `auto` value is currently not deciding correctly. With the fix in the value buckets I don't need to enforce using unwrapped unions, the automatic detection does it right.

mtth:
Thank you for the context.

This change will not help you since you are uploading Avro-encoded files to Snowflake. All unions have the same Avro encoding, independent of their in-memory representation. You will have to check on the Snowflake side to see how they represent the data.

W.r.t. the logic, a major concern is that it only works with values with a named constructor. No other values currently have this requirement.

joscha:
> You will have to check on the Snowflake side to see how they represent the data.

with `'never'` I seem to just get a plain JSON object. I haven't tried wrapped unions.

> W.r.t. the logic, a major concern is that it only works with values with a named constructor. No other values currently have this requirement.

Yes, hence the guard.

mtth:
> Yes, hence the guard.

It's not sufficient unfortunately. It would break many users of unwrapped unions, for example the common nullable case:

```javascript
const type = avro.Type.forSchema([
  'null',
  {type: 'record', name: 'test.Foo', fields: [{name: 'id', type: 'string'}]},
]);

type.toBuffer({id: 'abc'}); // Not valid anymore
```

(We should add a test for this.)

Another issue is that the library should be self-contained: any constructors required for use with unions should be included in--or generated by--the library. Record constructors are already generated but meant to be opt-in. They also aren't compatible with this implementation, which means that certain values don't roundtrip anymore:

```javascript
const val = type.fromBuffer(Buffer.from('0206616263', 'hex')); // Generated record instance
type.toBuffer(val); // Throws
```

For this feature to be worth adding, I think it needs to allow arbitrary record representations. Even if we assume the use of constructors to disambiguate branches, there is no obvious single way to name them. Another user would be justified in requesting that their own records, named slightly differently, also be supported.

Here is one way we could achieve this:

1. We allow the `unwrapUnions` option to also accept a function, which is expected to return a disambiguation function (e.g. `(types: ReadonlyArray<Type>) => ((val: unknown) => number) | undefined`).
2. This function is called on each union with its branches' types. If it returns a non-null function, it will be called each time a value's branch needs to be inferred and should return the branch's index. In this case the union will use an unwrapped representation. Otherwise, the union will be wrapped.

This single function gives users full flexibility to decide how to disambiguate unions. In your case, it could be a simple constructor check using the naming convention from the PR. Other users would be free to use their own convention, or non-constructor-based methods, or something else entirely.

WDYT?

joscha:
> > Yes, hence the guard.
> 
> It's not sufficient unfortunately. It would break many users of unwrapped unions, for example the common nullable case:

> Another issue is that the library should be self-contained: any constructors required for use with unions should be included in--or generated by--the library. Record constructors are already generated but meant to be opt-in. They also aren't compatible with this implementation, which means that certain values don't roundtrip anymore:

> For this feature to be worth adding, I think it needs to allow arbitrary record representations. Even if we assume the use of constructors to disambiguate branches, there is no obvious single way to name them. Another user would be justified in requesting that their own records, named slightly differently, also be supported.

Okay, I get your point, thank you for elaborating. Reading the constructor name is definitely not a great implementation option anyway.

> Here is one way we could achieve this:
> 
> 1. We allow the `unwrapUnions` option to also accept a function, which is expected to return a disambiguation function (e.g. `(types: ReadonlyArray<Type>) => ((val: unknown) => number) | undefined`) or nothing.
> 2. This function is called at schema parsing time on each union with its branches' types. If it returns a non-null (function) value, that function will be called each time a value's branch needs to be inferred and should return the branch's index. In this case the union will use an unwrapped representation. Otherwise, the union will be wrapped.
> 
> This single function gives users full flexibility to decide how to disambiguate unions. In your case, it could be a simple constructor check using the naming convention from the PR. Other users would be free to use their own constructor naming convention, or something else entirely.
> 
> WDYT?

I like it. This is what we would use instead of the result from `getValueBucket` in those cases when a projection function has been supplied AND returns a defined value, correct?

> (We should add a test for this.)

I can do this when I open the pull request for the above?

joscha:
I believe the above implementation suggestion can also solve https://github.com/mtth/avsc/pull/225#issuecomment-473723338 - you can simply return the discriminator of the union to ensure there is no wrapping needed.

mtth:
> I like it. This is what we would use instead of the result from getValueBucket in those cases when a projection function has been supplied AND returns a defined value, correct?

That's right.

> I can do this when I open the pull request for the above?

A separate tiny PR adding the test would be best.

joscha:
> A separate tiny PR adding the test would be best.

https://github.com/mtth/avsc/pull/477

joscha:
> That's right.

Changes are in

lib/types.js
```diff
@@ -197,7 +199,20 @@ class Type {
         return Type.forSchema(obj, opts);
       });
       if (!UnionType) {
-        UnionType = isAmbiguous(types) ? WrappedUnionType : UnwrappedUnionType;
+        // either automatic detection or we have a projection function
+        if (typeof opts.wrapUnions === 'function') {
+          // we have a projection function
+          try {
+          UnionType = typeof opts.wrapUnions(types) !== 'undefined'
```
joscha, line 206: we could potentially store the generated function here already, then there would only be one call to `wrapUnions` for each set of types instead of two, if we assume generating the projection function could possibly be expensive.

joscha:
@mtth PTAL

lib/types.js
mtth: Let's refactor the code to only call the function once per union. It's more efficient and less surprising for callers.

For example we can add a third argument to `UnwrappedUnionType` to forward it (or `false` if it did not return a function). When the argument is `undefined`, `UnwrappedUnionType` should still call `opts.wrapUnion` to support direct calls.

types/index.d.ts
```diff
@@ -96,6 +96,15 @@ interface EncoderOptions {
...
 }
 
+/**
+ * A projection function that is used when unwrapping unions.
+ * This function is called at schema parsing time on each union with its branches' types.
+ * If it returns a non-null (function) value, that function will be called each time a value's branch needs to be inferred and should return the branch's index.
+ * The index muss be a number between 0 and length-1 of the passed types.
+ * In this case (a branch index) the union will use an unwrapped representation. Otherwise (undefined), the union will be wrapped.
+ */
+type ProjectionFn = (types: ReadonlyArray<Type>) => ((val: unknown) => number) | undefined;
```
mtth, line 106: Can we rename to `BranchProjection`? Slightly more explicit and avoids introducing a new abbreviation.

Please also reformat the comment to fit within 80 characters.

lib/types.js
```diff
@@ -1251,13 +1266,31 @@ class UnwrappedUnionType extends UnionType {
 
     this._dynamicBranches = null;
     this._bucketIndices = {};
+
+    this.projectionFunction = (val) => this._bucketIndices[getValueBucket(val)];
```
mtth, line 1270: Let's consolidate logic to avoid unnecessary fields (e.g. `_bucketIndices` when a projection is present). For example something like:

```javascript
constructor (schema, opts, proj) {
  super(schema, opts);
  if (proj == null && typeof opts?.wrapUnions == 'function') {
    proj = opts.wrapUnions(this.types);
  }
  this._getIndex = proj ? projectionIndexer(proj) : defaultIndexer(this.types);
}
```

`defaultIndexer` encapsulates the current logic, moving it out of the constructor along with `_getBranchIndex`. `projectionIndexer` is lightweight, just adding error handling to the projection.

mtth (commented):
Thanks @joscha.

Will you also want this backported to `5.x`? (It will probably be at least a few weeks before `6.0` is released.)

joscha:
> Will you also want this backported to `5.x`? (It will probably be at least a few weeks before `6.0` is released.)

I wasn't going to - given it's not a fix, I think it could reasonably be expected to upgrade if the feature is needed? Do you think we need to?

types/index.d.ts
joscha: 040aed7

joscha: 040aed7

lib/types.js
joscha: changes in 510d272 - not sure I prefer all of it, as `projectionFn` is now a public interface to `UnwrappedUnionType`. We could possibly set it as a member on the instance of `UnionType`, less transparent but won't become a public interface.

When I tried to make `this._bucketIndices` local variables I noticed that this member is still used by `_copy` - I am unsure whether changes are needed in that function with the projection change?

lib/types.js
```diff
@@ -1226,6 +1243,44 @@ UnionType.prototype._branchConstructor = function () {
...
+  return (val) => {
+    const index = projectionFn(val);
+    if (typeof index !== 'number') {
+      throw new Error(`Projected index '${index}' is not valid`);
+    }
+    return index;
+  };
+}
+
+function generateDefaultIndexer() {
```
joscha, line 1257: this might be better as a private function on the class, given it still uses class context for `this._bucketIndices` and `this._getBranchIndex`.

lib/types.js
```diff
@@ -1246,38 +1301,33 @@ UnionType.prototype._branchConstructor = function () {
...
-  _getIndex (val) {
-    let index = this._bucketIndices[getValueBucket(val)];
-    if (this._dynamicBranches) {
-      // Slower path, we must run the value through all branches.
-      index = this._getBranchIndex(val, index);
-    }
-    return index;
-  }
-
   _getBranchIndex (any, index) {
```
joscha, line 1331: we could move this out as well or alternatively as described above, move the `defaultIndexer` into the class.

lib/types.js
mtth: The idea here is to move default indexing logic out of `UnwrappedUnion` into this indexer. We can move the method out and replace the fields with local variables. Something like:

```javascript
function generateDefaultIndexer(types) {
  let dynamicBranches = null;
  let bucketIndices = {};

  function getBranchIndex (any, index) {
    // Equivalent logic to _getBranchIndex...
  }

  types.forEach(function (type, index) {
    // Same as currently, replacing this._dynamicBranches with
    // dynamicBranches and similar for bucketIndices. 
  });

  return (val) => {
    // Same as in the forEach above here.
  };
}
```

We also need to perform a couple small changes in `_copy` to replace the accesses to `_bucketIndices`. For example:

```javascript
case 1: // Here
  if (isJsonBuffer(val)) {
    let bufIndex = this.types.findIndex(t => getTypeBucket(t) === 'buffer');
    if (bufIndex >= 0) {
      index = bufIndex;
    }
  }
  index ??= this._getIndex(val);
  break;
case 2:
  if (val === null) {
    index = this._getIndex(null); // And here
    // ...
```

We could optimize `case 1` to avoid `findIndex` calls but that can wait. We may remove the `coerceBuffers` option altogether as discussed [here](https://github.com/mtth/avsc/pull/452#issuecomment-2342308104).

lib/types.js
```diff
@@ -1246,38 +1301,33 @@ UnionType.prototype._branchConstructor = function () {
...
+  /**
+   *
+   * @param {*} schema
+   * @param {*} opts
+   * @param {Function|undefined} projectionFn The projection function used
+   *                                          to determine the bucket for the
+   *                                          Union. Falls back to generate
+   *                                          from `wrapUnions` parameter
+   *                                          if given.
+   */
```
mtth, line 1313: Let's delete this comment, the third argument is an implementation detail and should not be considered part of the public API.

lib/types.js
mtth: Yep, let's move it out. Carving unnecessary logic out makes `UnwrappedUnion` simpler and bucket inference more modular.

test/test_types.js
```diff
@@ -3505,6 +3505,54 @@ suite('types', () => {
...
+      };
+
+      // TODO: replace this with a mock when available
+      // currently we're on mocha without sinon
+      function mockWrapUnions() {
+        mockWrapUnions.calls = typeof mockWrapUnions.calls === 'undefined'
+          ? 1
+          : ++mockWrapUnions.calls;
+        return wrapUnions.apply(null, arguments);
+      }
```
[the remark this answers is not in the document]
joscha: This is exactly the idea of a spy - I was quite surprised a project this size does not have anything available, yet. It would make sense to add sinon I think, although longterm it might actually be better to ditch `mocha` and possibly got with the native node testing lib?

Anyway, changed in 5ecec5e

test/test_types.js
```diff
@@ -3505,6 +3505,54 @@ suite('types', () => {
...
+        return (animal) => {
+          const animalType = ((animal) => {
+            if ('bark' in animal) {
+              return 'Dog';
+            } else if ('meow' in animal) {
+              return 'Cat';
+            }
+            throw new Error('Unknown animal');
+          })(animal);
+          return types.indexOf(types.find(type => type.name === animalType));
```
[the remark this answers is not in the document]
joscha: This is the actual logic of the `wrapUnions` function, translating a given schema / name into a deterministic index? Are you suggesting to hardcode the index further up instead of first determining the Union type from the given object and then translating it into an index? If yes, I think that would not improve the test implementation as suddenly it relies of information that is not obvious anymore and prone to failing if `animalTypes` is changed?

types/index.d.ts
```diff
@@ -96,6 +96,17 @@ interface EncoderOptions {
   syncMarker: Buffer;
 }
 
+/**
+ * A projection function that is used when unwrapping unions.
+ * This function is called at schema parsing time on each union with its branches' types.
+ * If it returns a non-null (function) value, that function will be called each time a value's branch needs to be inferred and should return the branch's index.
+ * The index muss be a number between 0 and length-1 of the passed types.
+ * In this case (a branch index) the union will use an unwrapped representation. Otherwise (undefined), the union will be wrapped.
```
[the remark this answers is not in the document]
joscha: > Can you reformat to fit within 80 characters? Once this is merged, please remind me if I forget to update the API wiki.

Changed in 52873d0.
Sorry, not top of mind and this file also has signatures that go beyond the 80. An `.editorconfig` file would possibly make sense for this repo, if this is something that is expected everywhere?
Or even a `prettier`?

lib/types.js
```diff
@@ -196,11 +198,26 @@ class Type {
...
+          // we have a projection function
+          try {
+            projectionFn = opts.wrapUnions(types);
+            UnionType = typeof projectionFn !== 'undefined'
+              // projection function yields a function, we can use an Unwrapped type
+              ? UnwrappedUnionType
+              : WrappedUnionType;
+          } catch(e) {
+            throw new Error(`Error generating projection function: ${e}`);
+          }
```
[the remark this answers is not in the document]
joscha: I noticed whilst testing that it is quite confusing for a user if the `wrapUnions` function producing a function and the produced function itself errors in the same executed codepath, hence I error early here on the first call, but wanted to make sure someone using that public interface understands better what the issue is, as otherwise it reads like an internal problem.
I am not hung up on it though, if you feel we shouldn't wrap and remove context, we can not wrap.

test/test_types.js
```diff
@@ -3505,6 +3505,54 @@ suite('types', () => {
       assert(Type.isType(t.field('unwrapped').type, 'union:unwrapped'));
     });
 
+    test('union projection', () => {
```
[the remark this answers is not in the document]
joscha: > Can you add a test with a union function which returns `undefined`?

`wrapUnions` returning `undefined` as specified a valid to kick the serializer into auto mode or the projection function returning an invalid `undefined` with an error assertion?
I am assuming the former: 7629f6b

lib/types.js
```diff
@@ -196,11 +198,26 @@ class Type {
       let types = schema.map((obj) => {
         return Type.forSchema(obj, opts);
       });
+      let projectionFn;
       if (!UnionType) {
-        UnionType = isAmbiguous(types) ? WrappedUnionType : UnwrappedUnionType;
+        // either automatic detection or we have a projection function
+        if (typeof opts.wrapUnions === 'function') {
+          // we have a projection function
```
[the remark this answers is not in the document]
joscha: c0ead8b

lib/types.js
joscha: removed in 0e6c4cd; marked the param with an underscore prefix and a jsdoc-style inline comment, so it shows at least up in intellisense - I think the private parameter is confusing at best. I think writing it on the object before when the union type is determined or possibly using a `Symbol` in opts would be better.

lib/types.js
joscha: bd2c6dd

lib/types.js
joscha: bd2c6dd

lib/types.js
joscha: > When I tried to make `this._bucketIndices` local variables I noticed that this member is still used by `_copy` - I am unsure whether changes are needed in that function with the projection change?

changed in bd2c6dd

lib/types.js
mtth: It's not ideal, but better than mutating the options (which may be reused to surprising results, etc.). Once `avsc` is using TypeScript the constructor can be hidden from users of the library altogether.

lib/types.js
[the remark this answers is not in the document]
mtth: OK. My current instinct is that users will be able to debug this from the stack trace directly (which is truncated if we wrap, making the underlying cause harder to find).

Let's avoid wrapping for now. We can revisit if/when we introduce an error representation which fully supports causes.

lib/types.js
```diff
@@ -1226,6 +1241,60 @@ UnionType.prototype._branchConstructor = function () {
...
+    if (Type.isType(type, 'abstract', 'logical')) {
+      dynamicBranches.push({index, type});
+    } else {
+      let bucket = getTypeBucket(type);
+      if (bucketIndices[bucket] !== undefined) {
+        throw new Error(`ambiguous unwrapped union: ${j(this)}`);
+      }
+      bucketIndices[bucket] = index;
+    }
+  }, this);
```
mtth, line 1287: We can omit `this` now.

lib/types.js
```diff
@@ -1246,54 +1315,21 @@ UnionType.prototype._branchConstructor = function () {
...
-    let index = this._bucketIndices[getValueBucket(val)];
-    if (this._dynamicBranches) {
-      // Slower path, we must run the value through all branches.
-      index = this._getBranchIndex(val, index);
     }
-    return index;
-  }
+    this._getIndex = _projectionFn
+      ? generateProjectionIndexer(_projectionFn)
+      : generateDefaultIndexer.bind(this)(this.types);
```
mtth, line 1330: `generateDefaultIndexer(this.types)`

lib/types.js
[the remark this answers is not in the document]
joscha: 9f6eaffec66396cd891e39eca93a41e71cf8d321

lib/types.js
joscha: d6692e16e5593d8bd48778e336e07a8fd7b8820a

lib/types.js
joscha: d6692e16e5593d8bd48778e336e07a8fd7b8820a

mtth:
Thanks @joscha!

mtth:
> I think it could reasonably be expected to upgrade if the feature is needed? Do you think we need to?

I was checking in case you needed it to be backported. Otherwise, yes - no reason to.

joscha:
@mtth are you able to publish a new 6.x alpha version which includes this change by any chance, please?

mtth:
@joscha - done: [`6.0.0-alpha.14`](https://www.npmjs.com/package/avsc/v/6.0.0-alpha.14)

(Consider pinning, there will be breaking changes in later alpha releases.)

PR #474, opened 2024-09-13
Fix perf script

valadaptive:
Seems like avro.parse was removed in https://github.com/mtth/avsc/pull/428.


mtth:
Thanks @valadaptive.

Issue #468, opened 2024-09-10
Dynamic schema loading


joscha:
I am currently trying to load schemas from an automatically generated set of avro schemas.
It is basically a bunch `<typeName>.avsc` files in a directory.
They reference each other.
So when I load one like this:

```ts
Avro.Type.forSchema(JSON.parse(fs.readFileSync(__dirname + `/avro-schema/${fieldValueType}.avsc`, 'utf-8'))
```
where `fieldValueType` is `'CompanyValue'`:

```avsc
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "CompanyValue",
  "fields": [
    {
      "name": "type",
      "type": {
        "type": "enum",
        "name": "CompanyValue_type",
        "symbols": [
          "company"
        ]
      },
      "doc": "The type of value"
    },
    {
      "name": "data",
      "type": "model.CompanyData",
      "doc": ""
    }
  ]
}
```

it yields:
```
undefined type name: model.CompanyData
```
because the schema from `CompanyData.avsc` has not yet been loaded.

Initially I thought the answer would be to define an import hook: https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/files.js#L14
however it seems this is only supported for: https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/specs.js#L23

I can see that there is a local `registry` map I can pass into each `opt`: https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/types.js#L103 so I could just load all files into the registry with:
```ts
registry["model.CompanyData"] = Avro.Type.forSchema(JSON.parse(fs.readFileSync(...))
```
however I'd need to do that in a specific order which means I'd need to parse the model schemas myself beforehand.

I am wondering if you'd be up for adding an additional
```ts
resolver: (typeName: string) => Schema | Schema[]
```
to `opts`

which kicks in before https://github.com/mtth/avsc/blob/7cb76a6eb7b0d5b5a71c9cfb446acf3eff763006/lib/types.js#L159 to dynamically load a missing Schema (recursively).

joscha:
Workaround (not pretty but works):
```ts
class DynamicTypeRegistry {
  private readonly registry: Record<string, Avro.Type> = {};
  private readonly proxy = new Proxy(this.registry, {
[...]

PR #475, opened 2024-09-18
fix: `typeHook` can receive `string`

joscha:
`typeHook` is called for an unknown named type, e.g. `{ type: "record", name: "ns.MyName" }` yields `typeHook("ns.MyName", ...`

via https://github.com/mtth/avsc/issues/468#issuecomment-2357448344


mtth:
Thanks @joscha.

PR #476, opened 2024-09-23
fix: types

### LINKED ISSUE #461: Using seprately declared enum in union in record.
Hi,

I have a use case where I have an enum I want to declare outside the record. It works fine if I'm not using a union:

```
const enum = avro.Type.forSchema({
  name: "enum",
  type: "enum",
  symbols: ["foo", "bar", "baz"],
});

const record = avro.Type.forSchema({
  name: "record",
  type: "record",
  fields: [
    { name: "enum", type: enum },
  ],
});
```

But when I go to make it nullable with:

```
const record = avro.Type.forSchema({
  name: "record",
  type: "record",
  fields: [
    { name: "enum", type: ["null", enum], default: null },
  ],
});
```

I get a type error pointing to the `enum` in `["null", enum]`:
```
error TS2322: Type 'Type' is not assignable to type 'DefinedType'.
  Type 'Type' is not assignable to type 'FixedType & LogicalTypeExtension'.
    Type 'Type' is missing the following properties from type 'FixedType': type, size
```

If I inline the union it works fine... but I don't want to do that. It'd work for one of my unions, but another one is a union of a couple thousand entries, and I need it in multiple places so that'd get unwieldy fast.

I also tried using the name of the enum as a named type in the record, but that doesn't seem to work.

Thanks for the useful library!

joscha:
Fixes the type issue described in #461 and adds a fixture for it.
Also introduces a CI check to test the type definitions:

Screenshot 2024-09-23 at 12 37 31 PM

closes #461


.npmignore
```diff
@@ -0,0 +1 @@
+types/test/
```
joscha, line 1: this is so the tests don't get packaged in the resulting npm package.

package.json
```diff
@@ -43,18 +43,21 @@
...
   "scripts": {
+    "check-types": "tsc --strict --noEmit types/test/*.ts",
     "clean": "rm -rf coverage node_modules",
     "cover": "nyc mocha -- --ui tdd",
     "lint": "eslint etc/ lib/ test/ \"etc/scripts/**\"",
     "perf": "node --expose-gc ./etc/scripts/perf etc/schemas/*",
     "test": "mocha --ui tdd --reporter dot"
   },
   "devDependencies": {
+    "@types/node": "^22.5.5",
```
joscha, line 54: we could reduce this to the minimally supported version if we wanted, please let me know

types/test/461.ts
```diff
@@ -0,0 +1,15 @@
...
+    symbols: ["foo", "bar", "baz"],
+});
+
+const record = Avro.Type.forSchema({
+    name: "record",
+    type: "record",
+    fields: [
+        { name: "enum", type: ["null", e], default: null },
+    ],
+});
```
joscha, line 15:
```suggestion
});

```

types/index.d.ts
```diff
@@ -7,12 +7,11 @@
...
 
 import * as stream from 'stream';
-import { EventEmitter } from 'events';
 
 //"virtual" namespace (no JS, just types) for Avro Schema
 declare namespace schema {
-  export type AvroSchema = DefinedType | DefinedType[] | Type | Type[];
-  type DefinedType = PrimitiveType | ComplexType | LogicalType | string;
+  export type AvroSchema = DefinedType | DefinedType[];
+  type DefinedType = PrimitiveType | ComplexType | LogicalType | Type | string;
```
joscha, line 14: Without this change, the below yields:

```
tsc --strict --noEmit test/*.ts
test/461.ts:13:40 - error TS2322: Type 'Type' is not assignable to type 'DefinedType'.
  Type 'Type' is not assignable to type 'FixedType & LogicalTypeExtension'.
    Type 'Type' is missing the following properties from type 'FixedType': type, size

13         { name: "enum", type: ["null", e], default: null },
                                          ~


Found 1 error in test/461.ts:13
```

.npmignore
joscha, line 1:
```suggestion
types/test/

```

joscha, line 1: this is so the tests don't get packaged in the resulting npm package.

mtth: Can you update `files` in `package.json` instead?

package.json
mtth: This is fine.

.npmignore
joscha: Done in [fa198ce](https://github.com/mtth/avsc/pull/476/commits/fa198ce467728f9e37389799066b47f1829956a4) and 
8d0d137

joscha:
@mtth any chance to get this one in?

joscha:
> @joscha - please avoid bumping threads unless it's urgent or been more than a week, it adds noise to my notifications. I already do what I can to check in regularly.

Apologies, I did because part of this PR was overlapping with #479

PR #477, opened 2024-09-23
test: add test for synthetic constructor

joscha:
via https://github.com/mtth/avsc/pull/469#issuecomment-2357428899


joscha:
Screenshot 2024-09-23 at 4 28 14 PM

failed due to a slow checkout. Opened #478 to make sure it's not because we're using an outdated checkout plugin. Will update the branch after #478 has landed.

mtth:
Thanks @joscha.

Issue #423, opened 2022-12-27
Generate documentation and typings from JSDoc?


valadaptive:
Currently, the documentation seems to be manually maintained and updated alongside the code.

It may be beneficial to move the documentation alongside the code in the form of JSDoc comment blocks, using a tool like [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown) to generate documentation from it. This would make it easier to update the documentation to match the code, and there are even [linting tools](https://www.npmjs.com/package/eslint-plugin-jsdoc) that will ensure that the documentation is up-to-date. This would also make the documentation easier to access for people reading the code directly.

In addition, it's even possible to [generate TypeScript types from JSDoc](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html) automatically. This would ensure that typings are up-to-date in addition to the documentation.

joscha:
If we were to continue #479, then we could easily use https://typedoc.org/ to produce this. It has the ability to incorporate custom Markdown as well.

Issue #219, opened 2019-01-21
Types don't compile in node - Cannot find name 'Blob'.


tobalsgithub:
Ran into an issue compiling the type definitions in typescript in a node environment. I'm using this command `tsc --outDir bld --module commonjs --target ES2016` to compile and I hit this error

```
node_modules/avsc/types/index.d.ts(148,41): error TS2304: Cannot find name 'Blob'.
```

**Typescript: ** 2.9.2
**Node: **8.11.3

It appears that `Blob` is a global exposed on browsers but not available in a node environment. 

This issue discusses the same problem for a different library. https://github.com/DefinitelyTyped/DefinitelyTyped/issues/12044

I'd make a PR to fix this, but I don't actually know what the right solution is here. It feels like an anti-pattern to have the type definitions depend on something globally available only in certain environments, but I'm not well versed enough in typescript to know how to handle such a situation. 

For now, I've changed my compilation command to `tsc --outDir bld --module commonjs --target ES2016 --skipLibCheck` which gets me around the problem.

mtth:
Hey @tobalsgithub, I don't know enough about TypeScript typings to say what the best way forward is; from reading the issue you linked, it sounds like `--skipLibCheck` is a reasonable approach. Perhaps @andrew8er, @TysonAndre, or @reuzel (typing authors) have any ideas? I'm also curious why this issue has not popped up before, did something change recently?

andrew8er:
You may need to add `"dom"` to your `lib` section in your tsconfig.json:

```js
{
  "compilerOptions": {
    "lib": [
      "dom" // ...
     ]
  }
}
```

TysonAndre:
It's been a while.

I may have been using `avsc` from a project that added its own polyfill of Blob (e.g. sinon).

Or maybe I wasn't generating/validating the stubs in strict mode and didn't see that warning.

tobalsgithub:
@andrew8er, adding `dom` to my `lib` does in fact help, but that doesn't feel like the right thing to do. The challenge here is that `Blob` and other stuff from the `dom` really should not compile in node. They're not available in node, and thus allowing typescript to compile with them leads to unexpected runtime behavior.

It feels like the right solution here would be to somehow conditionally include the type definitions for  `createBlobEncoder` and `createBlobDecoder` depending on the environment. 

Here's a discussion that looks related. May just be something we have to workaround for the time being. https://github.com/Microsoft/TypeScript/issues/449

I appreciate y'all taking a look.

[...]

Issue #128, opened 2017-09-07
TypeScript Support


damonwildercarr:
Thanks for the hard work on this - great project.

Any plans to include type definitions for TypeScript?

mtth:
Typings would definitely be welcome. I doubt I'll have time to dig into TypeScript enough to add them anytime soon unfortunately; I'll keep this open in case someone else would like to take a stab at it. Thank you for the kind words and bringing this up!

andrew8er:
I have an incomplete version here: https://github.com/andrew8er/avsc/tree/feature/typescript-definitions

I am pretty sure there are some errors, since I'm only really using a small part of the API so far.

mtth:
Thanks @andrew8er! Can you submit a PR with what you have so far?

andrew8er:
https://github.com/mtth/avsc/pull/134

mtth:
Merged - thanks @andrew8er. Keeping this open since the typing file isn't complete yet.

TysonAndre:
What's the intent for the package https://www.npmjs.com/package/@avro/types ?

- I don't see it mentioned in this repo's issues
- Grepping the source code, I don't see "@avro/types"

Also, publishing declarations that are neither in "@types/avsc" nor packaged with "avsc" is non-standard and surprising. I proposed https://github.com/mtth/avsc/pull/146 ("bundling with your npm package") , but closed it since I'm not sure what the long term plan is.

According to https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

> Now that you have authored a declaration file following the steps of this guide, it is time to publish it to npm. There are two main ways you can publish your declaration files to npm:
> 
> 1. bundling with your npm package, or
> 2. publishing to the @types organization on npm.

mtth:
Hey @TysonAndre, thank you for the publishing tips. I'd created the `@avro/types` package to hold the _Avro_ type decoding/encoding logic from `avsc`: a long term effort to split `avsc` into modular components, unfortunately stalled for now. It isn't at all related to TypeScript so it sounds like the best option is to include the `types/` folder in this package.

yonkeltron:
Any status on this work? It's a really terrific initiative so I'm not nagging, just excited!

mtth:
@yonkeltron - still a WIP but we already have reasonably complete [typings](https://github.com/mtth/avsc/tree/master/types) thanks to @andrew8er and @TysonAndre.

[...]

PR #479, opened 2024-09-25
style: first start to move this package to TypeScript

joscha:
This adds typescript and transforms the smallest part (`platform.js`) to Typescript.
It keeps everything else the same, including locations and ES level. This allows TS and JS to co-exist for the time being.

Screenshot 2024-09-25 at 8 43 35 PM

via https://github.com/mtth/avsc/issues/461#issuecomment-2370016950

relates to: https://github.com/mtth/avsc/issues/423, https://github.com/mtth/avsc/issues/128, https://github.com/mtth/avsc/issues/219


tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
joscha, line 1: This is the smallest tsconfig possible for now (from https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html). Only addition is `strict`.

src/types.js
```diff
@@ -3015,7 +3015,7 @@ function isJsonBuffer(obj) {
  *
  * @param path {Array} Passed from hook, but unused (because empty where this
  * function is used, since we aren't keeping track of it for effiency).
- * @param val {...} The object to reject.
+ * @param val {any} The object to reject.
```
joscha, line 3018: This fails otherwise with:

```
src/types.js:3018:16 - error TS1014: A rest parameter must be last in a parameter list.

3018  * @param val {...} The object to reject.
                    ~~~


Found 1 error in src/types.js:3018
```

package.json
```diff
@@ -43,18 +43,22 @@
     "node": ">=6.0.0"
   },
   "scripts": {
+    "prepublish": "rimraf lib && tsc",
```
joscha, line 46: this generates `lib` on local install & before publish for now. Only difference is that `lib` is git ignored.

src/platform.ts
```diff
@@ -0,0 +1,20 @@
+import { createHash } from 'crypto';
+import { debuglog } from 'util'
```
joscha, line 2:
```suggestion
import { debuglog } from 'util';
```

lib/platform.js
```diff
@@ -1,21 +0,0 @@
-let crypto = require('crypto');
```
joscha, line 1: This one was first moved in 8483c05dcf82e552a29ac569812a12f7e63cba81 and then changed, however git doesn't show it as moved anymore due to the extension + content change.

src/platform.ts
```diff
@@ -0,0 +1,20 @@
+import { createHash } from 'crypto';
```
joscha, line 1: This generates to https://gist.github.com/joscha/b2f96866e4fb29c9dfadec7472ef0eb6 - you can see how close it is to the original `lib/platform.js`

tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "compilerOptions": {
+        "outDir": "./lib",
+        "allowJs": true,
+        "target": "es5",
```
valadaptive, line 5: Do we need to downlevel to ES5?

joscha: My intent with this pull request is to keep everything stable. The current JS code is ES5, so we should stay there.
Eventually we can adjust this.

joscha: Don't get me wrong, I think we should change this, but the current minimum node version defined for this package is 6.x, so there are some discussions to be had before we can make this a es2030 ;)

valadaptive: The code, at least in the master branch, uses ES6 features since https://github.com/mtth/avsc/pull/422 (classes, let/const, probably more).

joscha: You're right. I also just  checked https://node.green/ - even on 6.x (I doubt anyone would still be using it and I don't think we should support anything smaller than 18.x) we can use most features.

@mtth - are you opposed to making this `es6` - it would mean some exports, etc. would change and JS code generated from the typescript would be slightly different.

mtth: ES6 for `v6+` SGTM.

mtth:
@joscha - I wish you gave me a heads up before putting effort in. After thinking about it, I think this migration is probably best done by me: to streamline maintenance, I would want to use the same TypeScript standard setup as my other projects.

joscha:
> @joscha - I wish you gave me a heads up before putting effort in. After thinking about it, I think this migration is probably best done by me: to streamline maintenance, I would want to use the same TypeScript standard setup as my other projects.

I did in 

https://github.com/mtth/avsc/issues/461#issuecomment-2365245219

to which I got

https://github.com/mtth/avsc/issues/461#issuecomment-2370016950

and then

mtth:
Yep, sorry - I didn't mean to imply that creating a PR would be a good next step here.

joscha:
> Yep, sorry - I didn't mean to imply that creating a PR would be a good next step here.

Well you know, open source would do that. I also think transforming it step by step with TS and JS side by side would make this much easier and allow other people to contribute.

PR #483, opened 2024-10-02
fix: browser support without bundler/browserify

bill-min:
Check issue here: https://github.com/browserify/node-util/pull/62

It throws error `process is not defined` if bundler does not handle it when used in browser.

### DIFF
--- a/lib/platform.js
+++ b/lib/platform.js
@@ -1,5 +1,5 @@
 let crypto = require('crypto');
-let util = require('util');
+let util = typeof process !== "undefined" ? require('util') : { debuglog: console.log };
 
 /**
  * Compute a string's hash.


mtth:
`avsc` historically didn't support browser builds without a bundler. This would be a reasonable feature request, but we should do it cleanly (e.g. remove bundler related files and configurations).

bill-min:
If you check the node-util issue, ppl already complained that vite/webpack5/esbuild doesn't polyfill `process` anymore, which is what most ppl use nowadays. Its kinda weird that `util` sits under `browserify` but named `node-util` while the maintainer doesn't want to support browser compatibility at all...

bill-min:
Let me know what you think @mtth, I am happy to tweak the PR and/or refactor properly if we are aiming for the fix.

mtth:
I'm leaning towards removing `debuglog` altogether - would you mind updating the PR (`types.js` and the 2 `platform.js` files)?

bill-min:
#486

PR #486, opened 2024-10-07
fix: process is not defined in browser env

bill-min:
As discussed #483, `util` package does not support in browser env without proper bundler setup.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -172,7 +172,6 @@ class Type {
         try {
           return new DerivedType(schema, opts);
         } catch (err) {
-          console.error('failed to instantiate logical type for %s', schema.logicalType);
           if (opts.assertLogicalTypes) {
             // The spec mandates that we fall through to the underlying type if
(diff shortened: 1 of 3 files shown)


lib/types.js
```diff
@@ -172,10 +170,9 @@ class Type {
           registry[key] = opts.registry[key];
         });
         try {
-          debug('instantiating logical type for %s', schema.logicalType);
           return new DerivedType(schema, opts);
         } catch (err) {
-          debug('failed to instantiate logical type for %s', schema.logicalType);
+          console.error('failed to instantiate logical type for %s', schema.logicalType);
```
mtth, line 175: Let's delete this line too.

bill-min: Updated the PR, please check

mtth:
Thanks @bill-min.

bill-min:
Thanks @mtth for quick resolution!

PR #487, opened 2024-10-09
fix(UnwrappedUnionType): `this` reference

joscha:
Was broken via #469.

Without this fix, the error message is `undefined`:

Screenshot 2024-10-09 at 8 54 26 PM

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -1243,7 +1243,7 @@ function generateProjectionIndexer(projectionFn) {
   };
 }
 
-function generateDefaultIndexer(types) {
+function generateDefaultIndexer(types, self) {
   const dynamicBranches = [];
   const bucketIndices = {};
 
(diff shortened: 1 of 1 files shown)


lib/types.js
```diff
@@ -1314,7 +1314,7 @@ class UnwrappedUnionType extends UnionType {
     }
     this._getIndex = _projectionFn
       ? generateProjectionIndexer(_projectionFn)
-      : generateDefaultIndexer(this.types);
+      : generateDefaultIndexer(this.types, this);
```
joscha, line 1317: Alternative is:

```suggestion
      : generateDefaultIndexer.bind(this)();
```

and then use `this` and `this.types` directly.

mtth: Current is simpler.

mtth:
Thanks @joscha.

PR #488, opened 2024-10-17
fix(types): `Buffer` -> `Uint8Array`

joscha:
Via:

              > is it safe to assume that all occurrences of Buffer should become Uint8Array

Yes.

_Originally posted by @mtth in https://github.com/mtth/avsc/issues/452#issuecomment-2409033315_

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -74,7 +74,7 @@ type Schema = schema.AvroSchema;
 type Callback<V, Err = any> = (err: Err | null, value?: V) => void;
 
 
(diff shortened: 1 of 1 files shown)


joscha:
@mtth would you mind pushing a new 6.x with this change, please?

mtth:
@joscha - [`6.0.0-alpha.15`](https://www.npmjs.com/package/avsc/v/6.0.0-alpha.15)

joscha:
Thank you!

joscha:
We could catch issues like these by introducing https://github.com/tsdjs/tsd into this repository until such time that a conversion to TS (see #479) is finished.

PR #495, opened 2025-04-09
Fix link to spec in README.md

webmaster128:


mtth:
Thanks @webmaster128.

PR #508, opened 2026-06-25
fix: order StringType.compare by code point to match spec

spokodev:
## Problem

`StringType.compare()` and `compareBuffers()` disagree for the same string values:

```js
const type = avsc.Type.forSchema('string');
const a = '', b = '\u{1F600}'; // U+E000 vs U+1F600

type.compare(a, b);                                       // +1  (wrong)
type.compareBuffers(type.toBuffer(a), type.toBuffer(b));  // -1  (correct)
```

Both are public comparison APIs, so a caller cannot rely on a single sort order.

## Spec

The Avro spec [Sort order](https://avro.apache.org/docs/current/specification/#sort-order) defines string ordering as:

> strings are compared lexicographically by Unicode code point. Note that since UTF-8 is used as the encoding for strings, sorting the bytes (which is what `compareBuffers` does) gives the same result as comparing the code points.

So code-point order and UTF-8 byte order are identical, and `compareBuffers`/`_match` already follow it. `compare` did not.

## Root cause

`StringType` defines no `compare`, so it inherits `PrimitiveType.prototype.compare`, which is `utils.compare(a, b)` — a plain `a < b`. For strings, `<` compares by UTF-16 code unit, not code point. The two diverge for supplementary plane (astral) characters: a surrogate lead unit (`0xD800`–`0xDBFF`) sorts before BMP code points `U+E000`–`U+FFFF`, whereas by code point the astral character sorts after. Meanwhile `StringType._match` uses `tap1.matchBytes(tap2)` (UTF-8), so the two APIs disagree.

## Fix

Add `StringType.prototype.compare`, walking both strings by code point. This matches `_match`/`compareBuffers` and the spec. `RecordType` and `ArrayType` `compare` delegate to the field comparator, so they are corrected for string fields too.

## Verification

Red before, green after; added two tests in the `StringType` suite asserting `compare` agrees with `compareBuffers` and follows code-point order for astral characters. Full suite passes (504 tests).


spokodev:
Retargeted from `main` to `master`. It was opened against `main` (the schema-refactor branch) where it conflicted, but the fix is a v5-line change to `StringType._compare` and applies cleanly on `master` (the default branch, where comparable fixes like #488/#492/#495 landed). No code change, just the base.

